PR #90, opened 2019-12-29
Feature/webfront help index

RaidMax:


WebfrontCore/WebfrontCore.csproj
```diff
@@ -66,6 +66,7 @@
   <ItemGroup>
     <PackageReference Include="BuildBundlerMinifier" Version="3.2.435" />
     <PackageReference Include="BuildWebCompiler" Version="1.12.405" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="3.1.0" />
```
RaidMax: This is probably going to increase compressed size, but I need it for live reloading. Probably will see if there's a way to remove it during publish

PR #147, opened 2020-07-21
updated DefaultSettings.json

RaidMax:
updated DefaultSettings.json to add support for BO3 (T7) stock maps.


Application/DefaultSettings.json
```diff
@@ -728,6 +728,126 @@
         }
       ]
     },
+    "Game": "T7",
+      "Maps": [
```
RaidMax, line 732: invalid json here

mikzyy: is it still wrong? im just now seeing this

Bluscream, line 732: yes, just look at the entry below for reference

PR #182, opened 2021-01-02
Respecting admins and users privacy

sepehr-gh:
At RepZ we discovered we cant hide our iw4m panel address because it plays important role in our unban appeals, But on the other hand, we thought current panel doesn't respect privacy in some ways:

1. Normally admins mask in the game but players could easily find admins in iw4admin panel. Our fork fixes that as much as possible. Only admins (logged in users) can now see other admins in iw4admin panel.
2. We found that its some how against users privacy to let everyone see their message history (aka event history) so we decided to hide it.

1st was very important to us and we believe it maybe something useful for everyone, but 2nd one might be optional.
I understand the way I have fixed these issues may not be the best solution and I wish you implement these options if you reject my fix.

### DIFF
--- a/WebfrontCore/Views/Client/Find/Index.cshtml
+++ b/WebfrontCore/Views/Client/Find/Index.cshtml
@@ -20,7 +20,11 @@
                         <color-code value="@client.Name" allow="@ViewBag.EnableColorCodes"></color-code>
                     </a>
                 </div>
-                <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+                @if (ViewBag.Authorized){
+                    <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+                }else{
+                    <div class="col-4 level-color-0">User</div>
+                }
                 <div class="col-3 text-right">@client.LastConnectionText</div>
             </div>
         }
--- a/WebfrontCore/Views/Client/Find/Index.cshtml
+++ b/WebfrontCore/Views/Client/Find/Index.cshtml
@@ -45,7 +45,11 @@
                     <color-code value="@client.Name" allow="@ViewBag.EnableColorCodes"></color-code>
                 </a>
             </div>
-            <div class="p-2 level-color-@client.LevelInt">@client.Level</div>
+            @if (ViewBag.Authorized){
+                <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+            }else{
+                <div class="col-4 level-color-0">User</div>
+            }
             <div class="p-2 text-white-50">@client.LastConnectionText</div>
         </div>
     }
--- a/WebfrontCore/Views/Client/Privileged/Index.cshtml
+++ b/WebfrontCore/Views/Client/Privileged/Index.cshtml
@@ -26,6 +26,6 @@
     </div>
 }else{
     <div class="row">
-        <div class="bg-danger">This section only accessible for logged in users </div>
+        <div class="col-12 text-center bg-danger" style="padding: 10px">This section is only accessible for logged in users </div>
     </div>
 }
--- a/WebfrontCore/Views/Client/Privileged/Index.cshtml
+++ b/WebfrontCore/Views/Client/Privileged/Index.cshtml
@@ -2,7 +2,9 @@
 
 <h4 class="pb-3 text-center ">@ViewBag.Title</h4>
 
-<div class="row border-bottom">
+@if (ViewBag.Authorized)
+{
+    <div class="row border-bottom">
     @{
         foreach (var key in Model.Keys)
         {
@@ -21,4 +23,9 @@
             </div>
         }
     }
-</div>
+    </div>
+}else{
+    <div class="row">
+        <div class="bg-danger">This section only accessible for logged in users </div>
+    </div>
+}
--- a/WebfrontCore/Views/Client/Profile/Index.cshtml
+++ b/WebfrontCore/Views/Client/Profile/Index.cshtml
@@ -142,11 +142,14 @@
     </div>
 </div>
 
-<div class="row d-md-flex pt-2">
-    <div id="profile_events" class="text-muted text-left pl-4 pr-4 pl-md-0 pr-md-0">
-        @await Component.InvokeAsync("ProfileMetaList", new { clientId = Model.ClientId, count = 30, offset = 0, startAt = DateTime.UtcNow, metaType = Model.MetaFilterType })
+@if (ViewBag.Authorized)
+{
+    <div class="row d-md-flex pt-2">
+        <div id="profile_events" class="text-muted text-left pl-4 pr-4 pl-md-0 pr-md-0">
+            @await Component.InvokeAsync("ProfileMetaList", new { clientId = Model.ClientId, count = 30, offset = 0, startAt = DateTime.UtcNow, metaType = Model.MetaFilterType })
+        </div>
     </div>
-</div>
+}
 
 <div class="row">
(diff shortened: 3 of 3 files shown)


WebfrontCore/Views/Client/Find/Index.cshtml
```diff
@@ -20,7 +20,11 @@
                         <color-code value="@client.Name" allow="@ViewBag.EnableColorCodes"></color-code>
                     </a>
                 </div>
-                <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+                @if (ViewBag.Authorized){
+                    <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+                }else{
+                    <div class="col-4 level-color-0">User</div>
```
RaidMax, line 26: This `User` string should come from translations
`@ViewBag.Localization["GLOBAL_PERMISSION_USER"]`

WebfrontCore/Views/Client/Find/Index.cshtml
```diff
@@ -20,7 +20,11 @@
                         <color-code value="@client.Name" allow="@ViewBag.EnableColorCodes"></color-code>
                     </a>
                 </div>
-                <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+                @if (ViewBag.Authorized){
```
RaidMax, line 23: Should be configuration driven.
Add a new bool property to ApplicationConfiguration.cs called `EnableAdminPrivacy` or something like that. Then set that value on the viewbag in `BaseController.cs`.
This statement will then be something like: `@if (!ViewBag.Authorized && ViewBag.EnableAdminPrivacy) { // show fake level } else { // show actual level }} `

WebfrontCore/Views/Client/Find/Index.cshtml
```diff
@@ -41,7 +45,11 @@
                     <color-code value="@client.Name" allow="@ViewBag.EnableColorCodes"></color-code>
                 </a>
             </div>
-            <div class="p-2 level-color-@client.LevelInt">@client.Level</div>
+            @if (ViewBag.Authorized){
```
RaidMax, line 48: Configuration driven as well

WebfrontCore/Views/Client/Profile/Index.cshtml
```diff
@@ -74,9 +74,15 @@
         }
         else
         {
-            <div id="profile_level" class="font-weight-bold h4 mb-0 level-color-@Model.LevelInt">
-                @Model.Level
-            </div>
+            if(ViewBag.Authorized){
```
RaidMax, line 77: Configuration

WebfrontCore/Views/Client/Profile/Index.cshtml
```diff
@@ -74,9 +74,15 @@
...
+                <div id="profile_level" class="font-weight-bold h4 mb-0 level-color-@Model.LevelInt">
+                    @Model.Level
+                </div>
+            }else{
+                <div id="profile_level" class="font-weight-bold h4 mb-0 level-color-0">
+                    User
```
RaidMax, line 83: Translation

WebfrontCore/Views/Client/Profile/Index.cshtml
```diff
@@ -142,11 +148,14 @@
     </div>
 </div>
 
-<div class="row d-md-flex pt-2">
-    <div id="profile_events" class="text-muted text-left pl-4 pr-4 pl-md-0 pr-md-0">
-        @await Component.InvokeAsync("ProfileMetaList", new { clientId = Model.ClientId, count = 30, offset = 0, startAt = DateTime.UtcNow, metaType = Model.MetaFilterType })
+@if (ViewBag.Authorized)
```
RaidMax, line 151: Configuration.. but also sensitive events are already filtered out here...

WebfrontCore/Views/Client/Privileged/Index.cshtml
```diff
@@ -2,7 +2,9 @@
 
 <h4 class="pb-3 text-center ">@ViewBag.Title</h4>
 
-<div class="row border-bottom">
+@if (ViewBag.Authorized)
```
RaidMax, line 5: Is probably more effective to just conditionally add page in `_Layout.cshtml`
image
However incase someone tried to access the URL anyway, in `ClientController.cs->PrivilegedAsync`, you can return a 401 if they're not authorized

WebfrontCore/Views/Client/Privileged/Index.cshtml
```diff
@@ -21,4 +23,9 @@
             </div>
         }
     }
-</div>
\ No newline at end of file
+    </div>
+}else{
+    <div class="row">
```
RaidMax, line 28: This should be in the translation but if we follow my earlier suggestion, then it is no longer needed

WebfrontCore/Views/Client/Find/Index.cshtml
```diff
@@ -41,7 +45,11 @@
...
+                <div class="col-4 level-color-@client.LevelInt">@client.Level</div>
+            }else{
+                <div class="col-4 level-color-0">User</div>
```
RaidMax, line 51: Translation

RaidMax (changes requested):
I'm excited to have some community contributions! However I have a few issues with the suggested changes
1) I would argue that a list of admin names is not breach of privacy in any tangible way. The purpose of having an admin list is for players to identify and verify who the admins are when they need administrative assistance. If everyone is masked ingame and I'm unable to access an admin list of the webfront anyone can claim they're admin. Conversely I don't know who to look for ingame if I have an issue that requires admin assistance. In my opinion the admin list is a way to assert authoritative presence. The only realistic scenario I see for hiding admins is if you're trying to trick someone into cheating thinking that no admins are online. My argument for that though is if they don't cheat because an admin is online that's a pretty effective prevention of cheating by show of authoritative power.
2) For logged in users the only available data is chat message and non sensitive kick/ban info. This is a 1:1 correlation with the information publicly available in game (chat messages and kick/ban messages). For password protected servers all the chat messages are obfuscated. So current behavior mirrors the information you have access to by joining and playing on the server.

Despite my disagreement with some aspects of this, I'm ok with this being an opt-in functionality after making the suggested changes!

sepehr-gh:
I'm sorry to say I don't have knowledge on .Net and that is clearly the reason why I haven't done this in right ways.
Anyways, about the suggestions itself:

We had the same mindset as you when we first started to use iw4admin. But we got to the point that we realized people will change their behavior or turn of cheats when they know admins are around, and iw4admin panel was how they could easily find out about admins. Our argue is that there is a report command and if something is up in a server they can easily report. On the other hand, if any admin is available on the server at same time someone is cheating or abusing, no one needs to talk to the admin cause he/she is already aware of the situation. And if someone is in hurry to bring an admin to a server, "iw4admin panel admins list" doesn't help them at all! The way is to either report or jump into discord and say something.
With all of these in mind, if cheaters see admins in iw4admin, they just have to bookmark that admins profile and they will be aware of name changes as well... so "!mask" is basically useless.
Maybe this could all be optional, and I know it takes extra time and effort but it is important at least in our community.

And I agree with your point about chat messages being available to everyone. As I mentioned in first comment, its just optional thing in RepZ and we don't have a big reason behind it.

Thank you

Bluscream:
Intransparency is always gonna raise eyebrows, but i understand your issue with the admin list. Commands shouldn't be countered by panel features

RaidMax:
@sepehr-gh
Do you have enough information with the comments I’ve provided to make the changes?

sepehr-gh:
> @sepehr-gh
> Do you have enough information with the comments I’ve provided to make the changes?

I could understand only few of them so I'm afraid to say I wont be able to make all the changes.

PR #186, opened 2021-02-23
Added T7 aliases

Sparker-99:
1. T7 aliases by @mikzyy
2. Highrise for IW5 which is still in beta


mikzyy:
i forgot about this lol thank you

PR #196, opened 2021-04-07
Update Pluto T4 Parser

ChxseH:
Uses new static version string.


RaidMax (changes requested):
Let's bump up the version to `0.2` and you can add your name to authors ex: `'RaidMax, Chase`

ChxseH:
Bumped version.

RaidMax (approved):

PR #238, opened 2022-03-09
Hide IP information below specified permission

mjkzy:
TODO:
- Add configuration option for enabling this
- Add configuration option for what permission should be able to see it (ex: if specified `"Administrator"`, Administrator and above should be allowed to see it. everything below shouldn't)
- Review and check if i'm even using the whole `@if` and `if` stuff right

### DIFF
--- a/WebfrontCore/Views/Admin/_ListAuditLog.cshtml
+++ b/WebfrontCore/Views/Admin/_ListAuditLog.cshtml
@@ -6,6 +6,13 @@
 
 @foreach (var info in Model)
 {
+    // only let Administrator and above view Logged * actions (usually logging in or out)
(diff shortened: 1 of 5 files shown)


MrAmos123:
IIRC: The reason why Trusted+ can see IP information is that you should only be promoting people you trust.

mjkzy:
@MrAmos123 yeah i know but this is for the people who want it configurable. some people can be great moderators but i don't trust them beyond that. i wrote up a whole thing to Raid about it and he said he would accept a PR so i'm gonna finish it sometime.

RaidMax:
Appreciate the effort, but decided to implement it in a more extensible way in 91a05348f0763824c04f51858f43fbe0f1020ae4

Default permissions (allow all):
image

Allow only SeniorAdmin+ IP Address read

image

PR #257, opened 2022-08-08
Adding Mute for IW4x

Ayymoss:
Add support for !mute (!mu) to IW4x (and potentially other games)

Translation strings:
```
PLUGINS_MUTE_MUTED = "Muted"
PLUGINS_MUTE_UNMUTED = "Unmuted"
PLUGINS_MUTE_COMMANDS_MUTE_DESC = "Mute or unmute a target player"
```


Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+using SharedLibraryCore.Interfaces;
+
+namespace Mute.Commands;
+
+public class MuteCommand : Command
+{
+    public MuteCommand(CommandConfiguration config, ITranslationLookup layout) : base(config, layout)
+    {
+        Name = "mute";
+        Description = "Check your winnings!";
```
RaidMax: Nice copy paste. This should also use a translation key

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+        Alias = "mu";
+        Permission = EFClient.Permission.Moderator;
+        RequiresTarget = true;
+        Arguments = new[]
+        {
+            new CommandArgument
+            {
+                Name = "Player",
```
RaidMax: Use the translation key here

Plugins/Mute/MuteManager.cs
```diff
@@ -0,0 +1,20 @@
+﻿using SharedLibraryCore;
+
+namespace Mute;
+
+public class MuteManager
+{
+    public bool Mute(GameEvent gameEvent)
```
RaidMax: make this async

Plugins/Mute/MuteManager.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        if (gameEvent.Target.GetAdditionalProperty<bool>(Plugin.MuteKey))
+        {
+            gameEvent.Owner.ExecuteCommandAsync($"unmute {gameEvent.Target.ClientNumber}");
```
RaidMax: await

Plugins/Mute/MuteManager.cs
```diff
@@ -0,0 +1,20 @@
...
+            Plugin.DataManager.WritePersistentData(gameEvent.Target, false);
+            return false;
+        }
+
+        gameEvent.Owner.ExecuteCommandAsync($"muteClient {gameEvent.Target.ClientNumber}");
```
RaidMax: await

Ayymoss:
Should include all changes requested now. :)

Translation strings for sheet:
```
PLUGINS_MUTE_MUTED = "Muted"
PLUGINS_MUTE_UNMUTED = "Unmuted"
PLUGINS_MUTE_COMMANDS_MUTE_DESC = "Mute or unmute a target player"
```

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                Name = translationLookup["COMMANDS_ARGS_PLAYER"],
+                Required = true
+            }
+        };
+    }
+
+    public override async Task ExecuteAsync(GameEvent gameEvent)
+    {
+        var muteManager = new MuteManager();
```
RaidMax: Don't new this up every time the command is executed. Just initialize it once in the constructor as a private class field. Commands are singletons, ie every time the ExecuteAsync is called it's on the same command instance

Plugins/Mute/DataManager.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        _metaService = metaService;
+    }
+
+    private readonly IMetaServiceV2 _metaService;
+
+    public async void ReadPersistentData(EFClient client) => client.SetAdditionalProperty(Plugin.MuteKey,
+        bool.Parse((await _metaService.GetPersistentMeta(Plugin.MuteKey, client.ClientId))?.Value ?? "false"));
+
+    public async void WritePersistentData(EFClient client, bool state)
```
RaidMax: Async void should only be used for event callbacks. Use `async Task`

Plugins/Mute/MuteManager.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Mute;
+
+public class MuteManager
+{
+    public async Task<bool> Mute(GameEvent gameEvent)
+    {
+        if (gameEvent.Target.GetAdditionalProperty<bool>(Plugin.MuteKey))
+        {
+            await gameEvent.Owner.ExecuteCommandAsync($"unmute {gameEvent.Target.ClientNumber}");
+            Plugin.DataManager?.WritePersistentData(gameEvent.Target, false);
```
RaidMax: You should await this after addressing previous comment

Plugins/Mute/MuteManager.cs
```diff
@@ -0,0 +1,20 @@
...
+            return false;
+        }
+
+        await gameEvent.Owner.ExecuteCommandAsync($"muteClient {gameEvent.Target.ClientNumber}");
+        Plugin.DataManager?.WritePersistentData(gameEvent.Target, true);
```
RaidMax: await

Plugins/Mute/Plugin.cs
```diff
@@ -0,0 +1,55 @@
...
+        DataManager = new DataManager(metaService);
+    }
+
+    public string Name => "Mute";
+    public float Version => (float) Utilities.GetVersionAsDouble();
+    public string Author => "Amos";
+
+    public const string MuteKey = "IW4MMute";
+
+    public static DataManager? DataManager;
```
RaidMax: If you initialize this in the constructor it will never be null, remove the nullable qualifier

Plugins/Mute/Plugin.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    public Task OnEventAsync(GameEvent gameEvent, Server server)
+    {
+        if (server.GameName != Server.Game.IW4) return Task.CompletedTask;
+
+        switch (gameEvent.Type)
+        {
+            case GameEvent.EventType.Join:
+                DataManager?.ReadPersistentData(gameEvent.Origin);
```
RaidMax: After addressing the previous comment any place you reference data manager you can remove the null coalesce

Plugins/Mute/Plugin.cs
```diff
@@ -0,0 +1,55 @@
...
+    public float Version => (float) Utilities.GetVersionAsDouble();
+    public string Author => "Amos";
+
+    public const string MuteKey = "IW4MMute";
+
+    public static DataManager? DataManager;
+
+    public Task OnEventAsync(GameEvent gameEvent, Server server)
+    {
+        if (server.GameName != Server.Game.IW4) return Task.CompletedTask;
```
RaidMax: consider making a static field that has a collection of supported games;

```csharp

public readonly Game[] _supportedGames = new[] { Server.Game.IW4 };

//....
(!_supportedGames.Contains(server.GameName)

/....
```

You should also reference that in the mute command SupportedGames assignments

Plugins/Mute/Plugin.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        switch (gameEvent.Type)
+        {
+            case GameEvent.EventType.Join:
+                DataManager?.ReadPersistentData(gameEvent.Origin);
+
+                if (gameEvent.Origin.GetAdditionalProperty<bool>(MuteKey))
+                {
+                    server.ExecuteCommandAsync($"muteClient {gameEvent.Origin.ClientNumber}");
```
RaidMax: await 😢

Plugins/Mute/Plugin.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    public string Name => "Mute";
+    public float Version => (float) Utilities.GetVersionAsDouble();
+    public string Author => "Amos";
+
+    public const string MuteKey = "IW4MMute";
+
+    public static DataManager? DataManager;
+
+    public Task OnEventAsync(GameEvent gameEvent, Server server)
```
RaidMax: async

Plugins/Mute/Plugin.cs
```diff
@@ -0,0 +1,55 @@
...
+
+                if (gameEvent.Origin.GetAdditionalProperty<bool>(MuteKey))
+                {
+                    server.ExecuteCommandAsync($"muteClient {gameEvent.Origin.ClientNumber}");
+                }
+
+                break;
+        }
+
+        return Task.CompletedTask;
```
RaidMax: Won't need this in an async task

Plugins/Mute/Plugin.cs
RaidMax: just `return` in an async task

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+        Permission = EFClient.Permission.Moderator;
+        RequiresTarget = true;
+        Arguments = new[]
+        {
+            new CommandArgument
+            {
+                Name = translationLookup["COMMANDS_ARGS_PLAYER"],
+                Required = true
+            }
+        };
```
RaidMax, line 27: Per previous comment set supported games here otherwise default will be listed for all

RaidMax (changes requested):

Ayymoss:
Hold off reviewing the latest changes. I made quite a few changes to handling offline players better.

Ayymoss:
Alright, tested and all seems working fine. Please let me know any adjustments you'd want to be made.

Ayymoss:
Okay, please review when convenient.

Plugins/Mute/DataManager.cs
```diff
@@ -0,0 +1,36 @@
...
+    public DataManager(IMetaServiceV2 metaService)
+    {
+        _metaService = metaService;
+    }
+
+    private readonly IMetaServiceV2 _metaService;
+
+    public async Task<MuteState> ReadPersistentData(EFClient client)
+    {
+        var clientMeta = (await _metaService.GetPersistentMeta(Plugin.MuteKey, client.ClientId))?.Value ??
```
RaidMax: You should check first to see if it's in the additional  properties so we don't have to make another trip to the database. 
image
The most up to date value should be set when they join, so check the additional properties first, and if it's null then go to the database.

Plugins/Mute/DataManager.cs
```diff
@@ -0,0 +1,36 @@
...
+                         nameof(MuteState.Unmuted);
+        
+        return Parse<MuteState>(clientMeta);
+    }
+
+    public async Task WritePersistentData(EFClient client, MuteState state)
+    {
+        await _metaService.SetPersistentMeta(Plugin.MuteKey, state.ToString(), client.ClientId);
+        client.SetAdditionalProperty(Plugin.MuteKey, state);
```
RaidMax, line 29: This set of additional properties should arguably be done in the Mute manager, since based off the name of the class it should only deal with data storage/retrieval.

RaidMax (changes requested):

Plugins/Mute/DataManager.cs
Ayymoss: I've made a few changes to read for it either to make more or less sense.

RaidMax (approved):

PR #261, opened 2022-08-28
Fixes names longer than 34 characters cause of ^7.

efinst0rm:


RaidMax:
I'm going to go ahead and include this from my end as it's such a small change and I want to bundle it into the next release. I will add you as a co-author, even though I provided the original fix for you :P

PR #263, opened 2022-09-03
NoClip Fix - Removed NoClipOff - Toggle Hide

Ayymoss:
NoClip usage is now fixed, previously wouldn't correctly revert.
Made Hide toggleable to make usage more consistent throughout commands.
Removed obsolete NoClipOff & UnHide commands.
Fixed LockControls and made a toggle also

### DIFF
--- a/GameFiles/GameInterface/_integration_iw4x.gsc
+++ b/GameFiles/GameInterface/_integration_iw4x.gsc
@@ -56,7 +56,6 @@ RegisterClientCommands()
     scripts\_integration_base::AddClientCommand( "TakeWeapons",    true,  ::TakeWeaponsImpl );
     scripts\_integration_base::AddClientCommand( "SwitchTeams",    true,  ::TeamSwitchImpl );
     scripts\_integration_base::AddClientCommand( "Hide",           false, ::HideImpl );
-    scripts\_integration_base::AddClientCommand( "Unhide",         false, ::UnhideImpl );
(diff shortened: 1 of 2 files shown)


GameFiles/GameInterface/_integration_iw4x.gsc
```diff
@@ -330,71 +333,57 @@ NoClipImpl()
...
-
-    self IPrintLnBold( "NoClip enabled" );
 }
 
 HideImpl()
 {
     if ( !IsAlive( self ) )
     {
         self IPrintLnBold( "You are not alive" );
-        return;
```
RaidMax, line 353: why did we remove this return? Do we actually want to run the logic if the player is not alive?

Ayymoss: whoops

GameFiles/GameInterface/_integration_iw4x.gsc
```diff
@@ -330,19 +334,17 @@ NoClipImpl()
     else
     {
         self SetClientDvar( "sv_cheats", 1 );
```
diamante0018, line 336: I recall this not being necessary since the server can change cheat-protected dvars regardless of the value of sv_cheats (client side).
This logic applies because it's the host sending the command to the client.

RaidMax (approved):

PR #267, opened 2022-10-05
Extended functionality of Mute & Resolved PS IE Download Issue

Ayymoss:
Added additional muting commands.
Resolved errors in downloading translations and updating.

```
!unmute (!umu) <target> <reason> 
!tempmute (!tmu) <target> <duration> <reason> 
!listmutes (!lmu)
```

```
PLUGINS_MUTE_COMMANDS_LISTMUTES = (Color::Accent)- Currently Muted Players ({{count}}) -
PLUGINS_MUTE_COMMANDS_LISTMUTES_DESC = List active mutes
PLUGINS_MUTE_COMMANDS_LISTMUTES_EMPTY = There are no active mutes
PLUGINS_MUTE_COMMANDS_LISTMUTES_FORMAT = (Color::Accent)#{{index}} (Color::White)[(Color::Green)@{{clientId}}(Color::White), {{cleanedName}}, (Color::Yellow){{expiration}}(Color::White)] [(Color::Red)@{{adminId}}(Color::White), {{reason}}]

PLUGINS_MUTE_COMMANDS_MUTE_DESC = Mute or unmute a player
PLUGINS_MUTE_COMMANDS_MUTE_MUTED = Muted {{name}}
PLUGINS_MUTE_COMMANDS_MUTE_NOT_UNMUTED = {{name}} is already muted
PLUGINS_MUTE_COMMANDS_MUTE_TARGET_MUTED = (Color::Accent)You have been permanently muted for (Color::Yellow){{reason}}

PLUGINS_MUTE_COMMANDS_TEMPMUTE_BAD_FORMAT = Bad format! Expected <name|@clientId> <duration(m|h|d|w|y)> <reason>
PLUGINS_MUTE_COMMANDS_TEMPMUTE_DESC = Temporarily mute a target
PLUGINS_MUTE_COMMANDS_TEMPMUTE_TARGET_TEMPMUTED = (Color::Accent)You have been temporarily muted for (Color::Yellow){{expiration}} (Color::Accent)for (Color::Yellow){{reason}}
PLUGINS_MUTE_COMMANDS_TEMPMUTE_TEMPMUTED = Temporarily muted {{name}}

PLUGINS_MUTE_COMMANDS_UNMUTE_DESC = Unmute a target
PLUGINS_MUTE_COMMANDS_UNMUTE_NOT_MUTED = {{name}} is already unmuted
PLUGINS_MUTE_COMMANDS_UNMUTE_TARGET_UNMUTED = You have been unmuted for {{reason}}
PLUGINS_MUTE_COMMANDS_UNMUTE_UNMUTED = Unmuted {{name}}

PLUGINS_MUTE_REMAINING_TIME = Your mute will expire (Color::Yellow){{expiration}} (Color::White)((Color::Accent){{reason}}(Color::White))
PLUGINS_MUTE_NEVER = Never
PLUGINS_MUTE_EXPIRED = Mute expired
PLUGINS_MUTE_MIGRATED = IW4MAdmin Migrated
PLUGINS_MUTE_TARGET_EXPIRED = (Color::Green)Your mute has expired
PLUGINS_MUTE_UNMUTED = Unmuted

WEBFRONT_CLIENT_META_WAS_PENALIZED_MUTE = {{action->Permanently Muted}} by {{punisher}} for {{reason}}
WEBFRONT_CLIENT_META_PENALIZED_MUTE = {{action->Permanently Muted}} {{offender}} for {{reason}}

WEBFRONT_CLIENT_META_WAS_PENALIZED_TEMPMUTE = {{action->Temporarily Muted}} by {{punisher}} for {{reason}} with duration of {{time}}
WEBFRONT_CLIENT_META_PENALIZED_TEMPMUTE = {{action->Temporarily Muted}} {{offender}} for {{reason}} with duration of {{time}}

WEBFRONT_CLIENT_META_WAS_PENALIZED_UNMUTE = {{action->Unmuted}} by {{punisher}} for {{reason}}
WEBFRONT_CLIENT_META_PENALIZED_UNMUTE = {{action->Unmuted}} {{offender}} for {{reason}}

WEBFRONT_PENALTY_UNMUTE = "Unmute"
WEBFRONT_PENALTY_TEMPMUTE = "Temporary Mute"
WEBFRONT_PENALTY_MUTE = "Mute"
```
TODO: _interactionRegistration needs reason dialogue mapping.


Plugins/Mute/Commands/ListMutesCommand.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        if (!Plugin.MuteManager.HasMutes)
+        {
+            gameEvent.Origin.Tell(_translationLookup["PLUGINS_MUTE_COMMANDS_LISTMUTES_EMPTY"]);
+            return;
+        }
+        
+        var output = Plugin.MuteManager.Mutes.Select((x, index) =>
+            $"(Color::Yellow)#{index + 1}(Color::White): [(Color::Cyan)@{x.ClientId}(Color::White)] " +
+            $"(Color::White){x.CleanedName} [(Color::Red){(x.Expiration is not null ? x.Expiration.Value.HumanizeForCurrentCulture() : "Permanent")}" +
```
RaidMax: "Permanent" should use translation string.
Might be able to reuse WEBFRONT_ACTION_SELECTION_PERMANENT

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -23,20 +26,31 @@ public class MuteCommand : Command
...
         };
     }
-
-    private readonly MuteManager _muteManager = new();
-
+    
     public override async Task ExecuteAsync(GameEvent gameEvent)
     {
-        if (await _muteManager.Mute(gameEvent))
+        var match = Regex.Match(gameEvent.Data, BanRegex);
```
RaidMax: not sure why you're using regex here. 
Seems like all you're doing is just checking to make sure that the data has at least 1 character.
That's already handled by the command processing when the `Required` flag is true.

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -23,20 +26,31 @@ public class MuteCommand : Command
...
+        if (match.Success)
         {
-            gameEvent.Origin.Tell($"{_translationLookup["PLUGINS_MUTE_MUTED"]} {gameEvent.Target.Name}");
-            return;
-        }
+            var reason = match.Groups[1].ToString();
```
RaidMax: per previous, 'reason' will just be gameEvent.Data

Plugins/Mute/Commands/ListMutesCommand.cs
```diff
@@ -0,0 +1,38 @@
...
+    public override async Task ExecuteAsync(GameEvent gameEvent)
+    {
+        if (!Plugin.MuteManager.HasMutes)
+        {
+            gameEvent.Origin.Tell(_translationLookup["PLUGINS_MUTE_COMMANDS_LISTMUTES_EMPTY"]);
+            return;
+        }
+        
+        var output = Plugin.MuteManager.Mutes.Select((x, index) =>
+            $"(Color::Yellow)#{index + 1}(Color::White): [(Color::Cyan)@{x.ClientId}(Color::White)] " +
```
RaidMax: This should probably use the translation sheet

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -23,20 +26,31 @@ public class MuteCommand : Command
...
 
-        gameEvent.Origin.Tell($"{_translationLookup["PLUGINS_MUTE_UNMUTED"]} {gameEvent.Target.Name}");
+            if (await Plugin.MuteManager.Mute(gameEvent.Owner, gameEvent.Origin, gameEvent.Target, null, reason))
+            {
+                gameEvent.Origin.Tell($"{_translationLookup["PLUGINS_MUTE_MUTED"]} {gameEvent.Target.Name}");
```
RaidMax: This is incorrect/outdated usage of the translation. You're not guaranteed every language's word order will be the same.

Translation string (en-US): PLUGINS_MUTE_MUTED => "Muted {{clientName}}"
usage: `gameEvent.Origin.Tell(_translationLookup["PLUGINS_MUTE_MUTED"].FormatExt(gameEvent.Target.Name));`

This decouples the format from code and places it in the translation format. Then the FormatExt is responsible for plugging in the values where needed. It also makes more sense contextually on the translation sheet.

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -23,20 +26,31 @@ public class MuteCommand : Command
...
+                gameEvent.Target.Tell($"{_translationLookup["PLUGINS_MUTE_TARGET_MUTED"]}: {reason}");
```
RaidMax: Same comment as previous, also avoid colon

Plugins/Mute/Commands/MuteCommand.cs
```diff
@@ -23,20 +26,31 @@ public class MuteCommand : Command
...
+                return;
+            }
+
+            gameEvent.Origin.Tell($"{gameEvent.Target.Name} {_translationLookup["PLUGINS_MUTE_COMMANDS_MUTE_NOT_UNMUTED"]}");
```
RaidMax: same comment

Plugins/Mute/Commands/TempMuteCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+public class TempMuteCommand : Command
+{
+    private const string TempBanRegex = @"([0-9]+\w+)\ (.+)";
+
+    public TempMuteCommand(CommandConfiguration config, ITranslationLookup translationLookup) : base(config,
+        translationLookup)
+    {
+        Name = "tempmute";
+        Description = translationLookup["PLUGINS_MUTE_COMMANDS_TEMPMUTE_DESC"];
+        Alias = "tmu";
```
RaidMax: Why not just tm? Is it used elsewhere?

Plugins/Mute/Commands/TempMuteCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        var match = Regex.Match(gameEvent.Data, TempBanRegex);
+        if (match.Success)
+        {
+            var expiration = DateTime.UtcNow + match.Groups[1].ToString().ParseTimespan();
+            var reason = match.Groups[2].ToString();
+
+            if (await Plugin.MuteManager.Mute(gameEvent.Owner, gameEvent.Origin, gameEvent.Target, expiration, reason))
+            {
+                gameEvent.Origin.Tell($"{_translationLookup["PLUGINS_MUTE_COMMANDS_TEMPMUTE_TEMPMUTED"]} {gameEvent.Target.Name}");
```
RaidMax: Same translation comment

Plugins/Mute/Commands/TempMuteCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+                gameEvent.Target.Tell($"{_translationLookup["PLUGINS_MUTE_COMMANDS_TEMPMUTE_TARGET_TEMPMUTED"]} [{expiration.HumanizeForCurrentCulture()}]: {reason}");
```
RaidMax: same

Plugins/Mute/Commands/TempMuteCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+                return;
+            }
+
+            gameEvent.Origin.Tell($"{gameEvent.Target.Name} {_translationLookup["PLUGINS_MUTE_COMMANDS_MUTE_NOT_UNMUTED"]}");
```
RaidMax: same

Plugins/Mute/Commands/TempMuteCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+                Name = translationLookup["COMMANDS_ARGS_REASON"],
+                Required = true
+            }
+        };
+    }
+
+    public override async Task ExecuteAsync(GameEvent gameEvent)
+    {
+        var match = Regex.Match(gameEvent.Data, TempBanRegex);
+        if (match.Success)
```
RaidMax, line 46: If they type `!tmu player 5 reason` they get no output. Let them know if the duration format is incorrect

Plugins/Mute/Commands/UnmuteCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                Name = translationLookup["COMMANDS_ARGS_REASON"],
+                Required = true
+            }
+        };
+    }
+
+    public override async Task ExecuteAsync(GameEvent gameEvent)
+    {
+        var match = Regex.Match(gameEvent.Data, TempBanRegex);
```
RaidMax: unneeded check again because required is true

Plugins/Mute/Commands/UnmuteCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+        if (match.Success)
+        {
+            var reason = match.Groups[1].ToString();
+
+            if (await Plugin.MuteManager.Unmute(gameEvent.Owner, gameEvent.Origin, gameEvent.Target, reason))
+            {
+                gameEvent.Origin.Tell($"{_translationLookup["PLUGINS_MUTE_COMMANDS_UNMUTE_UNMUTED"]} {gameEvent.Target.Name}");
+                gameEvent.Target.Tell($"{_translationLookup["PLUGINS_MUTE_COMMANDS_UNMUTE_TARGET_UNMUTED"]}: {reason}");
```
RaidMax: fix translation to use formatext, avoid colon

Plugins/Mute/Commands/UnmuteCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+                return;
+            }
+
+            gameEvent.Origin.Tell($"{gameEvent.Target.Name} {_translationLookup["PLUGINS_MUTE_COMMANDS_UNMUTE_NOT_MUTED"]}");
```
RaidMax: fix translation

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,64 @@
...
+            {
+                ClientId = client.ClientId,
+                Reason = string.Empty,
+                Expiration = null,
+                AdminId = 0,
+                MuteState = tryParse ? clientMuteState : MuteState.Unmuted
+            };
+        }
+
+        client.SetAdditionalProperty(Plugin.MuteKey, clientMuteStateMeta);
```
RaidMax: I'm not a fan of this architecturally. The read method should not create a new instance. 
And honestly you should not have a meta state until you've either muted or unmuted someone.
There is nothing preventing you from performing the correct logic if the mute state is null. 
eg on connect if they don't have a mute state (null) you don't need to do anything.

Plugins/Mute/MuteManager.cs
```diff
@@ -1,21 +1,58 @@
...
 {
-    public async Task<bool> Mute(GameEvent gameEvent)
+    private List<MuteStateMeta> _muteList = null!;
+    public bool HasMutes => _muteList.Any();
+    public IEnumerable<MuteStateMeta> Mutes => _muteList;
+    public async void InitialiseMuteList() => _muteList = await Plugin.DataManager.ReadMuteList();
+
+    public async Task<bool> Mute(Server server, EFClient origin, EFClient target, DateTime? dateTime, string reason)
+    {
+        if ((await Plugin.DataManager.ReadPersistentData(target)).MuteState == MuteState.Muted) return false;
```
RaidMax: following the previous comment, that will make the result of read potentially null, so handle appropriately

RaidMax (changes requested):

Ayymoss:
Edit: You can review this now.

Note, if you're happy with the structure of ReadPersistentData, let me know. In the latest commit, ReadPersistentData will never return null, so will need to clean up other code and remove a bunch of null checks.

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+        var clientMuteStateMeta = client.GetAdditionalProperty<MuteStateMeta>(Plugin.MuteKey);
+        if (clientMuteStateMeta is not null) return clientMuteStateMeta;
+
+        // Get and return meta from database.
+        clientMuteStateMeta = await _metaService.GetPersistentMetaValue<MuteStateMeta>(Plugin.MuteKey,
+            client.ClientId);
+
+        if (clientMuteStateMeta is not null)
+        {
+            client.SetAdditionalProperty(Plugin.MuteKey, clientMuteStateMeta);
```
RaidMax: So I don't think the data manager should be responsible for setting the additional property here. Thing of this as retrieval only. The manager should be the place to set the additional property when needed

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+
+        // Get and return meta from old state.
+        var parseOldMuteState =
+            TryParse<MuteState>((await _metaService.GetPersistentMeta(Plugin.MuteKey, client.ClientId))?.Value,
+                out var oldClientMuteState);
+
+        // If still null, return client MuteStateMeta object. Not written.
+        // TODO: Ask Raid about this here... (Solves for nothing in DB and no old state (else a new joiner will trigger database read CONSTANTLY))
+        // TODO: If Raid agrees to keep this check, this Method will never return a nullable type.
+        if (!parseOldMuteState)
```
RaidMax: If both the new meta state and the old meta enum don't exist we should return null, not create a new instance. This method should only be for reading. There's too much logic embedded in this

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+            clientMuteStateMeta = new MuteStateMeta
+            {
+                ClientId = client.ClientId,
+                CleanedName = client.CleanedName,
+                Reason = string.Empty,
+                Expiration = DateTime.UtcNow,
+                AdminId = Utilities.IW4MAdminClient().ClientId,
+                MuteState = MuteState.Unmuted
+            };
+            client.SetAdditionalProperty(Plugin.MuteKey, clientMuteStateMeta);
```
RaidMax: Also set in the mute manager

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+            case MuteState.Unmuted:
+            default:
+                break;
+        }
+
+        // Return null or the populated meta.
+        return clientMuteStateMeta;
     }
 
-    public async Task WritePersistentData(EFClient client, MuteState state)
```
RaidMax: All this logic should be in the mute manager

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+    public async Task<MuteStateMeta> WritePersistentData(EFClient client, EFClient adminId, DateTime? dateTime,
+        MuteState state, string reason)
     {
-        await _metaService.SetPersistentMeta(Plugin.MuteKey, state.ToString(), client.ClientId);
-        client.SetAdditionalProperty(Plugin.MuteKey, state);
+        var clientMuteStateMeta = client.GetAdditionalProperty<MuteStateMeta>(Plugin.MuteKey) ??
```
RaidMax: The name of this method implies all data necessary to write should be provided as parameters. We should not need to look this up again before updating. Easiest thing to do is think of this doing a replacement of the value whether or not it exists already. Remember this is a state object we're writing so there should never really be a need for a partial update

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+        {
+            clientMuteStateMeta = new MuteStateMeta
+            {
+                ClientId = client.ClientId,
+                Reason = reason,
+                Expiration = dateTime,
+                AdminId = adminId.ClientId,
+                MuteState = state
+            };
+        }
```
RaidMax: Per previous comment this is redundant code because we're writing all the same properties regardless if it's being created or "updated"

Plugins/Mute/DataManager.cs
```diff
@@ -1,38 +1,119 @@
...
+
+        await _metaService.SetPersistentMetaValue(Plugin.MuteKey, clientMuteStateMeta, client.ClientId);
+        client.SetAdditionalProperty(Plugin.MuteKey, clientMuteStateMeta);
```
RaidMax: Don't need to set this here because in theory the data passed in should match the additional property instance exactly.

If needed this logic should be in mute manager

Plugins/Mute/MuteManager.cs
```diff
@@ -1,21 +1,67 @@
...
+
+    public async Task<bool> Mute(Server server, EFClient origin, EFClient target, DateTime? dateTime, string reason)
+    {
+        // Check if client has meta, if null, mute
+        var clientMetaState = await Plugin.DataManager.ReadPersistentData(target);
+        if (clientMetaState?.MuteState == MuteState.Muted) return false;
+
+        // Mute client
+        var client = server.GetClientsAsList().FirstOrDefault(x => x.NetworkId == target.NetworkId);
+        if (client is not null) await server.ExecuteCommandAsync($"muteClient {client.ClientNumber}");
```
RaidMax: This could throw an exception if the rcon connection fails. 
In that case the state would never be written.
Maybe that's ok, but consider the expected behavior. Maybe we want to store the fact they were muted even if the rcon connection fails, so the next time they join it tries to mute them again.

Plugins/Mute/MuteManager.cs
```diff
@@ -1,21 +1,67 @@
...
+    public async Task<bool> Unmute(Server server, EFClient origin, EFClient target, string reason)
     {
-        if (await Plugin.DataManager.ReadPersistentData(gameEvent.Target) == MuteState.Muted)
+        // Check if client has meta, if null, they can't be muted
+        var clientMetaState = await Plugin.DataManager.ReadPersistentData(target);
+        if (clientMetaState?.MuteState is MuteState.Unmuted or MuteState.Unmuting or null) return false;
+
+        // Unmute client
+        var client = server.GetClientsAsList().FirstOrDefault(x => x.NetworkId == target.NetworkId);
+        if (client is not null) await server.ExecuteCommandAsync($"unmute {client.ClientNumber}");
```
RaidMax: same comment about the exception behavior here

Ayymoss:
Everything should be updated to reflect the changes from the review. (Except for the commented one)

Plugins/Mute/Commands/ListMutesCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+    public override async Task ExecuteAsync(GameEvent gameEvent)
+    {
+        if (!Plugin.MuteManager.HasMutes)
+        {
+            gameEvent.Origin.Tell(_translationLookup["PLUGINS_MUTE_COMMANDS_LISTMUTES_EMPTY"]);
+            return;
+        }
+
+        var output = Plugin.MuteManager.Mutes
+            .Select((x, index) =>
```
RaidMax: nit: let's not use 'x' for selector var names. How about `mute`

Plugins/Mute/Commands/ListMutesCommand.cs
```diff
@@ -0,0 +1,40 @@
...
+                _translationLookup["PLUGINS_MUTE_COMMANDS_LISTMUTES_FORMAT"].FormatExt(index + 1, x.ClientId,
+                    x.CleanedName, x.Expiration is not null
+                        ? x.Expiration.Value.HumanizeForCurrentCulture()
+                        : _translationLookup["WEBFRONT_ACTION_SELECTION_PERMANENT"], x.AdminId, x.Reason));
+
+        gameEvent.Origin.Tell(_translationLookup["PLUGINS_MUTE_COMMANDS_LISTMUTES"]
```
RaidMax, line 36: Just a general thought. What happens when there's like 100 mutes? Entering this command will spam the chat for potentially a long time..

Plugins/Mute/MuteManager.cs
```diff
@@ -1,21 +1,173 @@
...
+    {
+        var clientMuteMeta = await ReadPersistentDataV2(client);
+        if (clientMuteMeta is not null) return clientMuteMeta;
+
+        // Return null if the client doesn't have old or new meta.
+        var muteState = await ReadPersistentDataV1(client);
+        clientMuteMeta = new MuteStateMeta
+        {
+            ClientId = client.ClientId,
+            CleanedName = client.CleanedName,
```
RaidMax, line 39: Is there a reason we're not retaining their color codes here and other locations? Webfront supports color codes if enabled, and the game of course will also.

Plugins/Mute/MuteManager.cs
```diff
@@ -1,21 +1,173 @@
...
-        await gameEvent.Owner.ExecuteCommandAsync($"muteClient {gameEvent.Target.ClientNumber}");
-        await Plugin.DataManager.WritePersistentData(gameEvent.Target, MuteState.Muted);
         return true;
     }
+
+    public async Task<bool> Unmute(Server server, EFClient origin, EFClient target, string reason)
+    {
+        var clientMuteMeta = await GetCurrentMuteState(target);
+        if (clientMuteMeta.MuteState is MuteState.Unmuted && clientMuteMeta.CommandExecuted) return false;
+        if (!target.IsIngame && clientMuteMeta.MuteState is MuteState.Unmuting) return false;
```
RaidMax, line 97: If muting from the webfront it may execute on a server the client isn't currently on. You should use the logic on like 115 to find the ingame client first.
eg: target = server.GetClientsAsList().FirstOrDefault(x => x.NetworkId == target.NetworkId) ?? target;

Then check the .IsIngame there

Plugins/Mute/MuteManager.cs
```diff
@@ -1,21 +1,173 @@
...
+        }
+
+        return clientMuteMeta;
+    }
+
+    public async Task<bool> Mute(Server server, EFClient origin, EFClient target, DateTime? dateTime, string reason)
     {
-        if (await Plugin.DataManager.ReadPersistentData(gameEvent.Target) == MuteState.Muted)
+        var clientMuteMeta = await GetCurrentMuteState(target);
+        if (clientMuteMeta.MuteState is MuteState.Muted && clientMuteMeta.CommandExecuted) return false;
```
RaidMax, line 69: how come we check if the client is ingame on the unmute, but not here?

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+
                         break;
                 }
 
                 break;
         }
     }
 
-    public Task OnLoadAsync(IManager manager)
+    private async Task<bool> IsExpiredMute(EFClient client, Server server)
```
RaidMax: It doesn't make sense for this method to both return the state of expiration and also unmute

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
     {
-        _interactionRegistration.RegisterInteraction(MuteInteraction, async (clientId, game, token) =>
+        var expiration = (await MuteManager.GetCurrentMuteState(client)).Expiration;
+
+        // Return if permanent mute and not yet expired
+        if (expiration is null) return false;
+        if (!(expiration < DateTime.UtcNow)) return false;
```
RaidMax: This logic should probably be in the mute manager

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+
+        await MuteManager.Unmute(server, Utilities.IW4MAdminClient(), client,
+            Utilities.CurrentLocalization.LocalizationIndex["PLUGINS_MUTE_EXPIRED"]);
```
RaidMax: Would recommend using the result of the call to an is expired method to then call the unmute method instead of doing it in the same place

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+
+        if (client.IsIngame)
         {
-            if (!clientId.HasValue || game.HasValue && !SupportedGames.Contains((Server.Game)game.Value))
-            {
-                return null;
-            }
+            client.Tell(Utilities.CurrentLocalization.LocalizationIndex["PLUGINS_MUTE_TARGET_EXPIRED"]);
```
RaidMax: Aren't you already letting them know they're unmuted in the mutemanager unmute implementation? This is going to be a bunch of text spammed

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
                         break;
                     case MuteState.Unmuting:
-                        await server.ExecuteCommandAsync($"unmute {gameEvent.Origin.ClientNumber}");
-                        await DataManager.WritePersistentData(gameEvent.Origin, MuteState.Unmuted);
                         break;
                     case MuteState.Unmuted:
+                        if (!muteMetaUpdate.CommandExecuted)
+                        {
+                            await server.ExecuteCommandAsync($"unmute {client.ClientNumber}");
+                            muteMetaUpdate.CommandExecuted = true;
```
RaidMax: You've now effectively duplicated the same work in two locations. (call to the unmute command)
I had in mind that you would call the mutemanger mute/unmute command and based on the state it could determine that it only needed to run the command and update the state.

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+
+                // Check if user has any meta set, else ignore (unmuted)
+                var muteMetaUpdate = await MuteManager.GetCurrentMuteState(client);
+                switch (muteMetaUpdate.MuteState)
+                {
+                    case MuteState.Muted:
+                        if (!muteMetaUpdate.CommandExecuted)
+                        {
+                            await server.ExecuteCommandAsync($"muteClient {client.ClientNumber}");
+                            muteMetaUpdate.CommandExecuted = true;
```
RaidMax: Same comment here

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+                        }
+
+                        await IsExpiredMute(client, server);
                         break;
                     case MuteState.Unmuting:
-                        await server.ExecuteCommandAsync($"unmute {gameEvent.Origin.ClientNumber}");
-                        await DataManager.WritePersistentData(gameEvent.Origin, MuteState.Unmuted);
```
RaidMax, line 42: If you're not using this case you can remove it.

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+                        break;
+                }
+
+                break;
+            case GameEvent.EventType.Say:
+                var muteMetaSay = await MuteManager.GetCurrentMuteState(gameEvent.Origin);
+                switch (muteMetaSay.MuteState)
+                {
+                    case MuteState.Muted:
+                        gameEvent.Origin.Tell(Utilities.CurrentLocalization
```
RaidMax, line 71: Does the chat even get sent to the server when they're muted?

Plugins/Mute/Plugin.cs
```diff
@@ -7,97 +7,203 @@ namespace Mute;
...
+                                    ? muteMetaJoin.Expiration.Value.HumanizeForCurrentCulture()
+                                    : Utilities.CurrentLocalization.LocalizationIndex["PLUGINS_MUTE_NEVER"],
+                                muteMetaJoin.Reason));
+
+                        // Handle unmute if expired.
+                        if (await IsExpiredMute(gameEvent.Origin, server)) break;
+
+                        // Handle mute of active mutes.
                         await server.ExecuteCommandAsync($"muteClient {gameEvent.Origin.ClientNumber}");
+                        muteMetaJoin.CommandExecuted = true;
```
RaidMax: I don't like that you've not got two places to maintain the mute command execution. I think this should be in the mute manager.

Plugins/Mute/MuteManager.cs
RaidMax: Oh also, how is this name being stored? If it's being stored at the time of mute and never updated that's problematic because they could change their name and it wouldn't make sense. However because this is a temporary measure until it's an official penalty type, I suppose this is fine.

Ayymoss:
F PR

PR #277, opened 2023-01-08
Update "tell" and "say" command

Chenterito:
This correctly displays "say_svName"


RaidMax:
Closing this PR as next release will support the custom say name for tell/say raw for games other than just IW4x.
Will be able to continue using the tell/say raw in the H1 parser and still get a custom say name.

PR #287, opened 2023-04-03
fix iw4x, integration. improve scripts overall

diamante0018:
- Fixes noclip in a guaranteed way it won't depend on iw4x updates breaking stuff (my fault lol)
- ~~Remove redundant threading on call to `setup` function: Init is already the specialized function for initializing stuff, threading a child thread should not be done~~
- Remove infinite call in loop to `notifyOnPlayerCommand`
- Fixes "isbot" on IW4x and PIW5 using specialized `call` GSC functionality to reference builtin pointers. (Other clients do not have this yet I think)


Ayymoss:
No problem for the heads up.

RaidMax:
Seems to be working fine in my testing, so going to merge this

PR #288, opened 2023-04-09
Fix Game Interface / AC Callbacks + support for T5ZM, T6MP and T6ZM for the Game Interface.

xerxes-at:
Fixes the following:

* On IW4, IW5 and T5 it was possible that `level.eventBus.gamename` was being set before `level.eventBus` was initialized as both happened through the init method in 2 different gsc files.

* Paths in `_integration_shared.gsc` pointing to `_integration_base.gsc` did not account for the different paths caused by deploy.bat moving the files to mp only folders for IW5 and T5.

* T6's _customcallbacks.gsc did not compile with the new compiler.

Adds the following:

* Support for T5ZM
    * Remove mp includes from `_integration_base`.
    * Define `GetXuid` as `overrideMethod` as T5ZM doesn't have it.
    * Define `GetPlayerFromClientNum` as `overrideMethod`  since getting all players is slightly different on T5ZM.
 * Support for T6 (MP and ZM)
    * Add `_integration_t6zm_helper.gsc` to support the same hud messages as MP.
    * Add `_integration_t6.gsc`

Other changes:

* Remove precompiled `_customcallbacks.gsc` for T6 as it can now load uncompiled GSC and the compiler used by Pluto is much better than the old one.
* Changes to deploy.bat to support the other changes.


diamante0018 (commented):
You should removed the compiled T6 plugin. There should not be a reason to encourage loading scripts that way.

GameFiles/GameInterface/_integration_iw4x.gsc
```diff
@@ -199,6 +199,11 @@ Log2Console( logLevel, message )
     PrintConsole( "[" + logLevel + "] " + message + "\n" );
 }
 
+_GetXUID()
```
diamante0018, line 204: No this is wrong. See how `ISTestClient` function is called in the same file.
You need to use the `call` keyword.

diamante0018 (changes requested):
You should removed the compiled T6 plugin. There should not be a reason t encourage loading scripts that way.

GameFiles/GameInterface/_integration_iw5.gsc
```diff
@@ -109,6 +109,11 @@ Log2Console( logLevel, message )
     Print( "[" + logLevel + "] " + message + "\n" );
 }
 
+_GetXUID()
```
diamante0018, line 114: Same as iw4 engine

GameFiles/GameInterface/_integration_t6.gsc
```diff
@@ -0,0 +1,569 @@
...
+        self Hide();
+        
+        self.isNoClipped = false;
+        
+        self IPrintLnBold( "NoClip disabled" );
+    }
+
+    self IPrintLnBold( "NoClip enabled" );*/
+
+    scripts\_integration_base::LogWarning( "NoClip is not supported on T6!" );
```
diamante0018, line 412: Noclip should be a built-in on t6 iirc

GameFiles/GameInterface/_integration_t6.gsc
```diff
@@ -0,0 +1,569 @@
...
+OnPlayerConnect()
+{
+    level endon ( "game_ended" );
+
+    for ( ;; )
+    {
+        level waittill( "connected", player );
+        
+        if ( scripts\_integration_base::_IsBot( player ) ) 
+        {
```
diamante0018, line 47: Use istestclient as it is a built-in

GameFiles/GameInterface/_integration_t5zm.gsc
```diff
@@ -0,0 +1,550 @@
...
+OnPlayerConnect()
+{
+    level endon ( "game_ended" );
+
+    for ( ;; )
+    {
+        level waittill( "connected", player );
+        
+        if ( scripts\_integration_base::_IsBot( player ) ) 
+        {
```
diamante0018, line 46: Same

GameFiles/GameInterface/_integration_t5.gsc
```diff
@@ -129,6 +129,11 @@ God()
     }
 }
 
+_GetXUID()
+{
```
diamante0018, line 135: On the 3arc engine you do not need the call keyword to call a built-in by reference but it should be attempted anyway

diamante0018 (approved):
Most of my comments can be discarded.
At the end because of the 3arc engine the call keyword can't be used so it forces us to use wrapper function.
I confirmed noclip is not supported on t5/t6 yet.
but I could be added to iw5 if it's not already done.

GameFiles/GameInterface/_integration_base.gsc
```diff
@@ -22,6 +20,8 @@ Setup()
     level.commonFunctions              = spawnstruct();
     level.commonFunctions.setDvar      = "SetDvarIfUninitialized";
     level.commonFunctions.isBot        = "IsBot";
+    level.commonFunctions.getXuid      = "GetXuid";
+    level.commonFunctions.GetPlayerFromClientNum = "GetPlayerFromClientNum";
```
RaidMax: nit: use camelCase for functionName
`level.commonFunctions.getPlayerFromClientNum`

GameFiles/GameInterface/_integration_base.gsc
```diff
@@ -360,7 +361,7 @@ GenerateJoinTeamString( isSpectator )
         }
     }
 
-    guid = self GetXuid();
+    guid = self [[level.overrideMethods[level.commonFunctions.getXuid]]]();
```
RaidMax: So after thinking about this, I think we should move the `GenerateJoinTeamString` into the _integration_shared.gsc 
I added this in here before I created the shared functionality file.
Philosophically the _integration_base,gsc should really only be responsible for facilitating the game interface data transfer/notifications. No direct game interaction or action on game events.

That should put the above registration of the overrideMethods more in line with what's already done in the shared gsc

RaidMax:
What's the reason for defining an entirely new set of methods for T5ZM? Could we not check a dvar to see if it's zm vs mp and then register the GetXuid and GetPlayerFromClientNum respectively?

Edit: Actually I just realized it might complain about referencing the method even if you aren't calling it.. Which if so is annoying.

GameFiles/GameInterface/_integration_t6.gsc
```diff
@@ -23,6 +23,8 @@ Setup()
     
     RegisterClientCommands();
     
+    _SetDvarIfUninitialized("sv_iw4madmin_autobalance",0);
```
diamante0018: minor complaint here but could you leave a space after the comma and also space out the parameters like in the rest of the file?

xerxes-at:
Your guess regarding t5zm is dead on, unlike on t6 it will complain about any and all mp scripts.

RaidMax (approved):

PR #306, opened 2023-06-04
gameinterface additions

INSANEMODE:
- add waittill_any_timeout overrideMethod to _integration_t6.gsc
- thread event handler calls in monitorEvents  in _integration_base.gsc

### DIFF
--- a/GameFiles/GameInterface/_integration_t6.gsc
+++ b/GameFiles/GameInterface/_integration_t6.gsc
@@ -27,6 +27,7 @@ Setup()
     level.overrideMethods[level.commonFunctions.waittillNotifyOrTimeout] = ::WaitillNotifyOrTimeoutWrapper;
     level.overrideMethods[level.commonFunctions.isBot]                   = ::IsBotWrapper;
     level.overrideMethods[level.commonFunctions.getXuid]                 = ::GetXuidWrapper;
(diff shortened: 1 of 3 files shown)


GameFiles/GameInterface/_integration_t6.gsc
```diff
@@ -27,6 +27,7 @@ Setup()
     level.overrideMethods[level.commonFunctions.waittillNotifyOrTimeout] = ::WaitillNotifyOrTimeoutWrapper;
     level.overrideMethods[level.commonFunctions.isBot]                   = ::IsBotWrapper;
     level.overrideMethods[level.commonFunctions.getXuid]                 = ::GetXuidWrapper;
+    level.overrideMethods[level.commonFunctions.waitTillAnyTimeout]       = common_scripts\utility::waittill_any_timeout;
```
RaidMax: You'll need to use the wrapper pattern here (see IsBotWrapper and GetXuidWrapper).
Not all CoDs support the `call` keyword so to avoid that incompatibility I do not use it in the _base or _shared code.
As a result all function pointers must be to a local function.

INSANEMODE: Addressed in b02a5f498f3eca88ebc9ebf60bfc459d0307b161
Also added to iw5, since the function also existed there.

Threading has been removed from event handlers, as well.

RaidMax (approved):

PR #311, opened 2023-10-12
Fix possible infinite loop exploit for t6 and iw5

ineedbots:
Possibility of an exploiter killing an enemy and disconnecting within the `wait` window to trigger an infinite loop.

IW4x gsc script have this fixed already.


RaidMax:
Thanks for the PR, was probably missed when the original iw4x issue was identified

PR #320, opened 2024-01-28
Iw6x Support

yfbsei:
`RegisterScriptCommand` // work
`RequestUrlObject` // work

```command = SpawnStruct(); 
command.property``` // not work

`RegisterScriptCommandObject` // not work

`set sv_iw4madmin_integration_enabled 1` in iw6x server.cfg breaks iw4m-admin or has no effect. Better left omitted.

### DIFF
--- a/Application/DefaultSettings.json
+++ b/Application/DefaultSettings.json
@@ -1252,6 +1252,10 @@
           "Name":  "zombie_coast"
         },
         {
+          "Alias": "Shangri-La",
+          "Name":  "zombie_temple"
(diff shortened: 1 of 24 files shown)


diamante0018 (approved):
I have tested this with the other two required game scripts and I can confirm it compiles without issues and has no runtime errors in the GSC VM.

I tested with the developer dvars turned on.

I will take the task upon myself to provide basic LTS for these game scripts for IW6x (now iw6-mod on AlterWare)

PR #321, opened 2024-02-02
Show() instead of Hide() If no clip disabled

realviolent:
It should set to Show(), otherwise you are still invisible if you use the command !noclip again.


RaidMax:
Good catch thanks

PR #344, opened 2024-08-22
Add /api/client/{clientId} Endpoint

Ayymoss:
This endpoint returns general information about the client ID provided.

image

It might be worth leaving this PR open for a bit for other members to suggest additional data that should return on this endpoint.


Steven-GC:
XUID please :)

Ayymoss:
> XUID please :)

We've spoken about this in tester, it makes more sense to return the NetworkID explicitly than transform it into the hex string.

RaidMax (approved):

PR #362, opened 2025-05-11
Utilise Minutes in TempBan for under 1 hour bans.

Ayymoss:
Previously, configurations with <1hr tempban period presets threw `int.Parse()` floating point issues as it converted it to <1. 

This resolves this by utilising minutes as the smallest time step.

### DIFF
--- a/SharedLibraryCore/Commands/NativeCommands.cs
+++ b/SharedLibraryCore/Commands/NativeCommands.cs
@@ -229,7 +229,7 @@ namespace SharedLibraryCore.Commands
     /// </summary>
     public class TempBanCommand : Command
     {
-        private static readonly string TempBanRegex = @"^([0-9]{1,5}\w+)\ (.+)";
+        private static readonly string TempBanRegex = @"^([0-9]{1,5}\p{L}+)\ (.+)";
         private readonly ApplicationConfiguration _appConfig;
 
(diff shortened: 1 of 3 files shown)


RaidMax (approved):

diamante0018:
This works but wouldn't it also be great if the code that threw the exception got refactored in a way that the underlying cause stops throwing under some circumstances? Perhaps I'm not understanding the code right now

Ayymoss:
@diamante0018 You're right. The underlying issues is still present. This fix was only for the webfront side. 

People could still use decimal in commands, but not sure why you would, but it's possible. :D

diamante0018:
I see. Ah I understand. Maybe I'll setup a server locally to check it out. Haven't wrote in c# in a while but if it's something I can do I'll make a pr

Ayymoss:
I can fix it if necessary, it's no issue. We should probably just reject the command or use the fallback value if the duration contains a decimal place.

Ayymoss:
Okay, this should input the issue too through regex constraints. If the regex can be bypassed, the `int.Parse()` issue will come back. But for all intents and purposes I think this is fine. 

Webfront now works with <1h durations.

Manual commands are now properly constrained.

Note, failing syntax will not return any message to the user.

PR #365, opened 2025-05-24
Add ResolvedExternalIPAddress to Server API responses

RaidMax:
This commit introduces the `ResolvedExternalIPAddress` property to the JSON responses for the server API endpoints located in `WebfrontCore/Controllers/API/Server.cs`.

The `ResolvedExternalIPAddress` property is nullable and contains the IPv4 string value of `Program.Manager.ExternalIPAddress` if the server's `ResolvedIpEndPoint.Address.IsInternal()` is true. Otherwise, it is null.

This enhances the server API by providing external IP information when a server is resolved to an internal IP address.


Ayymoss (approved):
LGTM

PR #367, opened 2025-05-28
EF Core LINQ Query Fails: DateTime.UtcNow - TimeSpan Cannot Be Transl…

Shorty:
Summary:
An unhandled exception occurs when running a LINQ query that includes DateTime.UtcNow - TimeSpan as a condition in Where(). EF Core cannot translate this expression into SQL.

> System.InvalidOperationException: The LINQ expression 'DbSet<EFAlias>()
>     .Where(e => e.LinkId == __linkId_0)
>     .Where(e => e.IPAddress != null)
>     .Where(e => e.DateAdded >= DateTime.UtcNow - ___appConfig_RecentAliasIpLinkTimeLimit_1)' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. 

Code causing the issue:
```
var recentlyUsedIps = await context.Aliases
    .Where(alias => alias.LinkId == linkId)
    .Where(alias => alias.IPAddress != null)
    .Where(alias => alias.DateAdded >= DateTime.UtcNow - _appConfig.RecentAliasIpLinkTimeLimit)
    .Select(alias => alias.IPAddress)
    .ToListAsync();
```

Fix:

```
var cutoffDate = DateTime.UtcNow - _appConfig.RecentAliasIpLinkTimeLimit;

var recentlyUsedIps = await context.Aliases
    .Where(alias => alias.LinkId == linkId)
    .Where(alias => alias.IPAddress != null)
    .Where(alias => alias.DateAdded >= cutoffDate)
    .Select(alias => alias.IPAddress)
    .ToListAsync();
```

Config Value:
`RecentAliasIpLinkTimeLimit = 7.0:00:00`

### DIFF
--- a/SharedLibraryCore/Services/PenaltyService.cs
+++ b/SharedLibraryCore/Services/PenaltyService.cs
@@ -198,10 +198,14 @@ namespace SharedLibraryCore.Services
         {
             await using var context = _contextFactory.CreateContext(false);
 
-            var recentlyUsedIps = await context.Aliases.Where(alias => alias.LinkId == linkId)
+            var cutoffDate = DateTime.UtcNow - _appConfig.RecentAliasIpLinkTimeLimit;
+
+            var recentlyUsedIps = await context.Aliases
+                .Where(alias => alias.LinkId == linkId)
                 .Where(alias => alias.IPAddress != null)
-                .Where(alias => alias.DateAdded >= DateTime.UtcNow - _appConfig.RecentAliasIpLinkTimeLimit)
-                .Select(alias => alias.IPAddress).ToListAsync();
+                .Where(alias => alias.DateAdded >= cutoffDate)
+                .Select(alias => alias.IPAddress)
+                .ToListAsync();
 
             if (!recentlyUsedIps.Any())
             {


RaidMax:
Is this an issue with an underlying SQL provider? Are you using MySQL/MariaDB?

This particular scenario was a limitation with an older version of Entity Framework Core, but was fixed in .NET 6 I think.

Shorty:
Thanks for your response, we're using SQLite.

Useful information regarding .NET 6, in hindsight, this should've probably been raised as an issue to identify a resolution.

Ayymoss:
Were you experiencing any crashes? Most use SQLite, but we've not seen any reports that would indicate this fix is necessary. There's likely a few more locations that need a similar change, if so.

Shorty:
No crashes, but plenty of errors in the logs.

If you want to close this PR and perhaps will raise as an issue to further discuss.

RaidMax:
I was able to investigate this and validate the concern. It looks to be a limitation with the SQLite provider's LINQ to entity translation. I've checked the code base for any more occurrences, and this is the only scenario where this occurs. 
There are two other similar datetime comparisons, but they are contained in a static expression, which seems to behave differently.

This fix follows the same approach as all the other datetime comparisons.
I'm surprised you're getting notable amount of logging as this is relatively infrequent edge case for ban evasion detection.
Thanks for opening this PR and bringing it to my attention!

RaidMax (approved):

PR #368, opened 2025-05-31
Fix improper reference to app config for default BaseEventParser

RaidMax:


Ayymoss (approved):
Hmm... LGTM keep up the great work!

PR #386, opened 2026-01-10
Handle Stale Servers

Ayymoss:
Some games are shit, and their crash behaviour keeps RCON open and responds as if nothing is wrong. This is how we can try to detect it.

Fix your games. smh

### DIFF
--- a/Application/IW4MServer.cs
+++ b/Application/IW4MServer.cs
@@ -49,13 +49,7 @@ namespace IW4MAdmin
         private const int REPORT_FLAG_COUNT = 4;
         private long lastGameTime = 0;
 
-        // Fail-state detection tracking
-        private DateTime? _lastScoreActivity = DateTime.UtcNow;
-        private DateTime? _lastJoinLeaveActivity = DateTime.UtcNow;
-        private DateTime? _lastStatsActivity = DateTime.UtcNow;
-        private bool _isFailState = false;
-        private DateTime? _failStateDetectedAt = null;
-        private Dictionary<long, int> _previousPlayerScores = new Dictionary<long, int>();
+        private readonly IServerStateChecker _stateChecker;
 
         private readonly IServiceProvider _serviceProvider;
         private readonly IClientNoticeMessageFormatter _messageFormatter;
@@ -65,7 +59,7 @@ namespace IW4MAdmin
         private readonly StatManager _statManager;
         private readonly ApplicationConfiguration _appConfig;
 
-        public override bool IsFailState => _isFailState;
+        public override bool IsErrorState => _stateChecker.IsErrorState(GetClientsAsList().Count(c => !c.IsBot));
 
         public IW4MServer(
             ServerConfiguration serverConfiguration,
@@ -74,7 +68,8 @@ namespace IW4MAdmin
             IMetaServiceV2 metaService, 
             IServiceProvider serviceProvider,
             IClientNoticeMessageFormatter messageFormatter,
-            ILookupCache<EFServer> serverCache) : base(serviceProvider.GetRequiredService<ILogger<Server>>(), 
+            ILookupCache<EFServer> serverCache,
+            IServerStateChecker stateChecker) : base(serviceProvider.GetRequiredService<ILogger<Server>>(), 
 #pragma warning disable CS0612
             serviceProvider.GetRequiredService<SharedLibraryCore.Interfaces.ILogger>(), 
 #pragma warning restore CS0612
@@ -91,6 +86,7 @@ namespace IW4MAdmin
             _commandConfiguration = commandConfiguration;
             _statManager = serviceProvider.GetRequiredService<StatManager>();
             _appConfig =  serviceProvider.GetService<ApplicationConfiguration>();
+            _stateChecker = stateChecker;
             
             IGameServerEventSubscriptions.MonitoringStarted += async (gameEvent, token) =>
             {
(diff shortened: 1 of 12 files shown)


Application/IW4MServer.cs
```diff
@@ -91,6 +98,35 @@ public IW4MServer(
                 await EnsureServerAdded();
                 await _statManager.EnsureServerAdded(gameEvent.Server, token);
             };
+            
+            // Subscribe to stats events for fail-state detection
+            IGameEventSubscriptions.ClientKilled += async (killEvent, token) =>
```
RaidMax: These subscriptions should done on the ServerDataCollector

Application/IW4MServer.cs
```diff
@@ -931,6 +942,13 @@ private async Task OnClientUpdate(EFClient origin)
             }
 
             client.Ping = origin.Ping;
+            
+            // Track score changes for fail-state detection
+            if (client.Score != origin.Score)
```
RaidMax: Not sure why we still have this here. This method is triggered from ClientUpdateEvent which can be subscribed to in the state checker and moved out of here entirely

Application/IW4MServer.cs
```diff
@@ -998,13 +1016,23 @@ async Task<List<EFClient>[]> PollPlayersAsync(CancellationToken token)
             UpdateHostname(statusResponse.Hostname);
             UpdateMaxPlayers(statusResponse.MaxClients);
 
+            // Track join/leave activity
+            if (connectingClients.Any() || disconnectingClients.Any())
```
RaidMax: This is determinable in the state tracker via client connect/disconnect events. Shouldn't need this here

Application/IW4MServer.cs
```diff
@@ -1146,6 +1174,26 @@ public override async Task<bool> ProcessUpdatesAsync(CancellationToken token)
                         return true;
                     }
 
+                    // If fail-state was just detected, disconnect all players
+                    if (_stateChecker.FailStateDetectedAt.HasValue &&
```
RaidMax: This conditional logic should be encapsulated in the _stateChecker class. This magic "30 seconds" should be handled by the _stateChecker as it may be different depending on client quirks

Application/IW4MServer.cs
```diff
@@ -1146,6 +1174,26 @@ public override async Task<bool> ProcessUpdatesAsync(CancellationToken token)
                         return true;
                     }
 
+                    // If fail-state was just detected, disconnect all players
+                    if (_stateChecker.FailStateDetectedAt.HasValue && 
+                        (DateTime.UtcNow - _stateChecker.FailStateDetectedAt.Value).TotalSeconds < 30)
+                    {
+                        // Disconnect all non-bot players to prevent false playtime accumulation
+                        var allClients = GetClientsAsList().Where(c => !c.IsBot && c.State == ClientState.Connected).ToList();
```
RaidMax: Why not disconnect all players? If the server is detected as a fail state the bots are also not "playing"

If the server enters a fail state while a player is in the process of connecting we should probably disconnect them as well. I don't think we need this filter at all

Application/Misc/ServerStateChecker.cs
```diff
@@ -0,0 +1,308 @@
...
+    /// </summary>
+    public void RecordActivity()
+    {
+        LastActivity = DateTime.UtcNow;
+    }
+
+    /// <summary>
+    /// Computes whether the server is in a fail-state based on configuration and player count
+    /// </summary>
+    public bool IsErrorState(int nonBotPlayerCount)
```
RaidMax: need to find a way to not require passing in the bot player count.

Application/IW4MServer.cs
```diff
@@ -90,6 +96,9 @@ public IW4MServer(
 
                 await EnsureServerAdded();
                 await _statManager.EnsureServerAdded(gameEvent.Server, token);
+                
+                // Initialize the state checker for fail-state detection (subscribes to events)
+                _stateChecker.Initialize(this);
```
RaidMax, line 101: With dynamic server functionality we need to make _stateChecker disposable and dispose of it so the event subscriptions are un-registered

SharedLibraryCore/Interfaces/IServerStateChecker.cs
```diff
@@ -0,0 +1,26 @@
...
+    
+    DateTime? LastActivity { get; }
+    DateTime? FailStateDetectedAt { get; }
+    
+    /// <summary>
+    /// Initialize the state checker for a specific server and subscribe to events
+    /// </summary>
+    void Initialize(Server server);
+    
+    void RecordActivity();
```
RaidMax: Based on the other comments this shouldn't need to be public

SharedLibraryCore/Interfaces/IServerStateChecker.cs
```diff
@@ -0,0 +1,26 @@
...
+    bool IsErrorState(int nonBotPlayerCount);
+    void ProcessPlayerUpdates(IEnumerable<EFClient> updatedClients, IEnumerable<EFClient> polledClients);
```
RaidMax: Same

SharedLibraryCore/Interfaces/IServerStateChecker.cs
```diff
@@ -0,0 +1,26 @@
...
+    bool CheckAndUpdateFailState(List<EFClient> currentClients);
+    TimeSpan? GetFailStateDuration();
```
RaidMax, line 34: Same

RaidMax (changes requested):

Application/Misc/ServerStateChecker.cs
RaidMax: Server is passed in on initialize, so we have access to the instance

RaidMax (approved):

PR #387, opened 2026-01-10
More CS Script Changes

Ayymoss:
Support for Commands, Webfront Interactions, and the slim ScriptPluginSettings.json for easy integration.


Ayymoss:
No longer needed.

PR #390, opened 2026-01-11
Fix 'Command' Support for CS 'Script' Plugins

Ayymoss:
Fixed Command Usage & Dropped Slim Config Support

### DIFF
--- a/Application/Plugin/CSharpScript/CsPluginCommandRegistrar.cs
+++ b/Application/Plugin/CSharpScript/CsPluginCommandRegistrar.cs
@@ -11,7 +11,7 @@ namespace IW4MAdmin.Application.Plugin.CSharpScript;
 /// <summary>
 /// Handles discovery, registration, and unregistration of commands from plugin assemblies.
 /// </summary>
-public class CsPluginCommandRegistrar(IServiceProvider rootServiceProvider, ILogger<CsPluginCommandRegistrar> logger)
+public class CsPluginCommandRegistrar(ILogger<CsPluginCommandRegistrar> logger, IManager manager)
 {
     /// <summary>
     /// Discovers Command classes from the plugin assembly and registers them with the manager.
@@ -22,13 +22,12 @@ public class CsPluginCommandRegistrar(IServiceProvider rootServiceProvider, ILog
 
         try
         {
-            var manager = rootServiceProvider.GetRequiredService<IManager>();
-
-            var commandTypes = assembly.GetTypes()
+            var commandTypes = assembly.GetExportedTypes()
                 .Where(type =>
                     type.IsClass &&
                     !type.IsAbstract &&
                     type.IsSubclassOf(typeof(Command)) &&
+                    // Ensure we don't accidentally pick up base commands from the core library
                     (type.Namespace == null || !type.Namespace.StartsWith(nameof(SharedLibraryCore))))
                 .ToList();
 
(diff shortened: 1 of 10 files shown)


Application/Plugin/CSharpScript/CsPluginServiceHost.cs
```diff
@@ -495,33 +496,30 @@ private async Task RegisterPluginCommands(CsPluginInstance instance, System.Refl
                 try
                 {
                     // Create a temporary command instance to get its name/alias
-                    var tempCommand = (Command)ActivatorUtilities.CreateInstance(
-                        _rootServiceProvider,
-                        commandType,
-                        commandConfig,
-                        translationLookup);
-                    
+                    var tempCommand = (Command)ActivatorUtilities.CreateInstance(instance.PluginServiceProvider, commandType);
```
RaidMax: There's some duplicated code in here. 
This loops through the command types twice and first time instantiates a temp command just to check to see if it exists already, then loops through them again and does what appears to be essentially the same thing.
I can't think of any reason why we can't just loop through the command types and do the instantiation and check all in one place.

RaidMax: Also there's a third iteration of this loop login in the `catch (InvalidOperationException ex) when (ex.Message.Contains("Duplicate command"))`

If we've removed previous instances of the command and there's still a duplicate we shouldn't be retrying anything, that's a valid conflict at the plugin code needs to be corrected a retry won't fix anything. That whole catch statement is unecessary.

Application/Plugin/CSharpScript/CsPluginFileWatcher.cs
```diff
@@ -0,0 +1,270 @@
...
+    {
+        _logger.LogInformation("Plugin removed: {FileName}", e.Name);
+
+        _ = Task.Run(async () =>
+        {
+            var semaphore = GetOrCreateFileLock(e.FullPath);
+            await semaphore.WaitAsync();
+            try
+            {
+                if (FileDeleted != null)
```
RaidMax: Missing removal from _lastChangeTime, minor memory leak

Application/Plugin/CSharpScript/CsPluginFileWatcher.cs
```diff
@@ -0,0 +1,270 @@
...
+            finally
+            {
+                semaphore.Release();
+            }
+        });
+    }
+
+    private SemaphoreSlim GetOrCreateFileLock(string filePath)
+    {
+        lock (_lockObject)
```
RaidMax: This is the wrong lock object. You should be locking _fileOperationLocks

RaidMax: Alternatively keep this as is and change the lock in CleanupFilePath

Application/Plugin/CSharpScript/CsPluginServiceHost.cs
```diff
@@ -109,198 +98,42 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
-    /// <summary>
-    /// Unified handler for Created and Changed events with proper debouncing.
-    /// Both events fire when copying a file, so we need to handle them identically.
-    /// </summary>
-    private void HandleFileEventAsync(string filePath, FileEventType eventType)
-    {
-        if (!ShouldProcessChange(filePath))
+        // Check if plugin is already loaded
+        bool isAlreadyLoaded;
+        lock (_lock)
```
RaidMax: This extra lock and contains key is redundant because UnloadPluginsAsync uses the lock and returns early if the key is not found. With this here you're locking twice for no reason

Application/Plugin/CSharpScript/CsPluginCommandRegistrar.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        try
+        {
+            var manager = rootServiceProvider.GetRequiredService<IManager>();
+
+            var commandTypes = assembly.GetTypes()
+                .Where(type =>
+                    type.IsClass &&
+                    !type.IsAbstract &&
+                    type.BaseType == typeof(Command) &&
```
RaidMax: Nit: if someone writes a helper "BaseCommand" for shared logic for a custom command collection and inherits it'll be skipped. May make sense to do `type.IsSubclassOf(typeof(Command))`
This was likely pulled from the binary plugin logic but doesn't hurt to improve

RaidMax (approved):

PR #396, opened 2026-02-03
Update Command NextMap

K-Faktor:
Added sv_maprotationcurrent check
when using the !nm command, the map was not detected correctly, but it is now fixed and works well

### DIFF
--- a/SharedLibraryCore/Commands/NativeCommands.cs
+++ b/SharedLibraryCore/Commands/NativeCommands.cs
@@ -1187,15 +1187,37 @@ namespace SharedLibraryCore.Commands
 
         public static async Task<string> GetNextMap(Server s, ITranslationLookup lookup)
         {
+            // FIRST: Try sv_maprotationcurrent (GSManager style)
+            var currentRotation = await s.GetDvarAsync<string>("sv_maprotationcurrent", token: s.Manager.CancellationToken);
+    
+            if (!string.IsNullOrEmpty(currentRotation?.Value))
+            {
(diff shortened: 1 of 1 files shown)


SharedLibraryCore/Commands/NativeCommands.cs
```diff
@@ -1187,31 +1187,57 @@ public NextMapCommand(CommandConfiguration config, ITranslationLookup translatio
...
         public static async Task<string> GetNextMap(Server s, ITranslationLookup lookup)
         {
+            // FIRST: Try sv_maprotationcurrent (GSManager style)
+            var currentRotation = await s.GetDvarAsync<string>("sv_maprotationcurrent", token: s.Manager.CancellationToken);
+    
+            if (!string.IsNullOrEmpty(currentRotation?.Value))
+            {
+                // Parse format: "gametype sd map mp_bloc ..."
+                var pattern = @"^\s*(?:gametype\s+([a-z]+)\s+)?map\s+([a-z0-9_.-]+)\s*(.*)$";
+                var match = Regex.Match(currentRotation.Value.ToLower(), pattern, RegexOptions.IgnoreCase);
```
diamante0018, line 1197: Wether or not to use regex to parse the map rotation is a question I'll leave to the maintainers
(the clients do not waste resources using regex: see iw4x-client map rotation.cpp, it's just easier to split the string by spaces and read the first token pair for map <string>).
Anyhow, you are already ignoring the case with the regex function so calling tolower() is a wasteful operation.

PR #397, opened 2026-02-10
Playtime Sparkline but for All Servers

Ayymoss:
How fun!

image


Application/Misc/ServerDataViewer.cs
```diff
@@ -216,5 +219,49 @@ public async Task<int> RankedClientsCountAsync(long? serverId = null, Cancellati
...
             }
         }
+
+        public async Task<ServerActivitySparklineResult> GetServerActivityAsync(Reference.Game? gameCode = null, int days = 30, CancellationToken token = default)
+        {
+            _serverActivityCache.SetCacheItem(async (set, ids, cancellationToken) =>
+            {
+                Reference.Game? game = ids != null && ids.Any() ? (Reference.Game?)ids.First() : null;
+
+                var thirtyDaysAgo = DateTime.UtcNow.Date.AddDays(-(days - 1));
```
RaidMax: nit: this won't always be 30 days if a non default parameter value is specified. How about `oldestDataConsidered `or something?

Application/Misc/ServerDataViewer.cs
```diff
@@ -216,5 +219,49 @@ public async Task<int> RankedClientsCountAsync(long? serverId = null, Cancellati
...
+                };
+
+                for (var i = 0; i < days; i++)
+                {
+                    var date = DateTime.UtcNow.Date.AddDays(-(days - 1) + i);
+                    result.DailyPlayTimeMinutes[i] = dailyData.TryGetValue(date, out var minutes) ? minutes : 0;
+                    result.TotalPlaytimeMinutes += (long)result.DailyPlayTimeMinutes[i];
+                }
+
+                return result;
```
RaidMax: I'm not certain, but could this be converted to use LINQ instead of an ugly for loop?

Data/Models/Server/EFGameStatistic.cs
```diff
@@ -0,0 +1,19 @@
...
+
+/// <summary>
+/// Game-level daily statistics (e.g. aggregated playtime). Per-server stats use <see cref="EFServerStatistics"/>.
+/// </summary>
+public class EFGameStatistic : SharedEntity
+{
+    [Key]
+    public long Id { get; set; }
+    /// <summary>Matches <see cref="Reference.Game"/>; stored as int for EF.</summary>
+    public int? GameName { get; set; }
```
RaidMax: I can't remember is there a reason we can't use the `enum` here? I would imagine it would automatically convert to the integral type.

SharedLibraryCore/Dtos/ServerActivitySparklineResult.cs
```diff
@@ -0,0 +1,7 @@
+namespace SharedLibraryCore.Dtos;
+
+public class ServerActivitySparklineResult
+{
+    public double[] DailyPlayTimeMinutes { get; set; } = Array.Empty<double>();
```
RaidMax: nit: I think lists are typically recommended over arrays for almost every use case any more (especially with .net 10)

WebfrontCore/Components/UI/Display/ActivitySparklineGraph.razor
```diff
@@ -0,0 +1,57 @@
...
+                    <div class="w-full rounded-t transition-all hover:opacity-80 @colorClass"
+                         style="height: @(height)px;">
+                    </div>
+                </Tooltip>
+            }
+        }
+    </div>
+</div>
+
+@code {
```
RaidMax: Move to code behind?

WebfrontCore/Components/Features/Home/Components/ServerActivitySparkline.razor.cs
```diff
@@ -0,0 +1,53 @@
+using Humanizer;
+using Microsoft.AspNetCore.Components;
+using WebfrontCore.Core.Services;
+
+namespace WebfrontCore.Components.Features.Home.Components;
+
+public partial class ServerActivitySparkline : ComponentBase
```
RaidMax, line 6: Would it make sense to inherit from `ActivitySparklineGraph` component for the shared functionality? I'm not super familiar with what the best practice is for component inheritance in blazor, but it seems like things like the AppState, DataService etc.. could be injected in the `ActivitySparklineGraph` component and referenced here and in the `ClientActivitySparklineGraph` as well to cut down on duplication

WebfrontCore/Components/UI/Display/ActivitySparklineGraph.razor
```diff
@@ -0,0 +1,57 @@
...
+                var height = value > 0 ? Math.Max(4, (int)(value / maxValue * 32)) : 2;
+                var colorClass = value > 0 ? "bg-primary" : "bg-line";
+                var playTime = TimeSpan.FromMinutes(value);
+                var tooltipText = value > 0
+                    ? $"{FormatPlaytime(playTime)} • {date:MMM d}"
+                    : $"{AppState.Loc("WEBFRONT_PROFILE_NO_PLAYTIME")} • {date:MMM d}";
+
+                <Tooltip Text="@tooltipText" WrapperClass="flex-1 flex items-end">
+                    <div class="w-full rounded-t transition-all hover:opacity-80 @colorClass"
+                         style="height: @(height)px;">
```
RaidMax, line 36: nit: ensure inline style is actually needed

WebfrontCore/WebfrontCore.csproj
```diff
@@ -46,6 +46,10 @@
         <PackageReference Include="GoogleAuthenticator" Version="3.2.0" />
         <PackageReference Include="Microsoft.AspNetCore.ConcurrencyLimiter" Version="10.0.1" />
         <PackageReference Include="Microsoft.Web.LibraryManager.Build" Version="3.0.71" />
+        <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="9.0.0">
```
RaidMax: Shouldn't be needed as none of the migrations should be dependent on the webfront core project for schema or startup

Application/Extensions/StartupExtensions.cs
```diff
@@ -14,17 +14,18 @@
...
+    private static ILogger _defaultLogger;
+    private static readonly LoggingLevelSwitch LevelSwitch = new();
+    private static readonly LoggingLevelSwitch MicrosoftLevelSwitch = new();
+    private static readonly LoggingLevelSwitch SystemLevelSwitch = new();
 
-        public static IServiceCollection AddBaseLogger(this IServiceCollection services,
-            ApplicationConfiguration appConfig)
+    extension(IServiceCollection services)
+    {
+        public IServiceCollection AddBaseLogger(ApplicationConfiguration appConfig)
```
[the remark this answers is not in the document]
RaidMax: Supported by current .net/c# lang release

WebfrontCore/Components/Features/Home/Components/ServerActivitySparkline.razor.cs
```diff
@@ -0,0 +1,53 @@
...
+            await LoadActivityDataAsync();
+        }
+    }
+
+    private async Task LoadActivityDataAsync()
+    {
+        try
+        {
+            var result = await DataService.GetServerActivitySparklineAsync(Game, CancellationToken.None);
+
```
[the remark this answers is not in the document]
RaidMax: It's fine

Application/Misc/ServerDataCollector.cs
```diff
@@ -17,127 +17,346 @@
...
+
+        var endDate = DateTime.UtcNow;
+        var startDate = endDate.Date.AddDays(-1);
+
+        var pruneDate = endDate.AddDays(-31);
+        try
+        {
+            var oldRecords = await context.GameStatistics
+                .Where(x => x.Date < pruneDate)
+                .ToListAsync(token);
```
RaidMax: Better to use `ExecuteDeleteAsync` for bulk remove instead of retrieve and remove range

Application/Misc/ServerDataCollector.cs
```diff
@@ -17,127 +17,346 @@
...
+            if (events.Count == 0) continue;
+
+            var distinctClientIdsInEvents = events.Select(e => e.ClientId).Distinct().ToList();
+            var clientRows = await context.Clients
+                .Where(c => distinctClientIdsInEvents.Contains(c.ClientId))
+                .Include(c => c.CurrentAlias)
+                .Select(c => new { c.ClientId, c.NetworkId, Name = c.CurrentAlias != null ? c.CurrentAlias.Name : null })
+                .ToListAsync(token);
+            var botClientIds = clientRows
+                .Where(c => c.NetworkId == (c.Name ?? "").StripColors().GenerateGuidFromString())
```
RaidMax: Can't we use .IsBot?

WebfrontCore/Components/Features/Home/Components/ServerActivitySparkline.razor.cs
Ayymoss: ActivitySparklineGraph is pure display (Data + TotalPlaytime only). The two wrappers do different data work (single DTO vs. paginated + session-pairing), so injecting DataService/Logger into the display would couple rendering to loading and block reuse when the caller already has data. Shared surface is ~2 fields + error fallback. Can extract a loader base class if you'd prefer, but would keep the display standalone.

PR #398, opened 2026-03-21
Add per-server RCON and log pipeline latency monitoring

Ayymoss:
## Summary
  - Adds per-server latency metrics (RCON round-trip time and log pipeline latency) measured via EMA smoothing
  - RCON RTT is measured automatically for all servers by timing existing status polls
  - Log pipeline latency requires the GSC Game Interface and works by sending probe dvars that the GSC echoes back via LogPrint, measuring the full pipeline delay
  - Exposes metrics via `IGameServer.LatencyMetrics` for plugin consumers and via the `/api/server` endpoint

  ## Changes
  - **SharedLibraryCore**: `ServerLatencyMetrics` EMA container, `LatencyProbeScriptEvent`, `LastRtt` on `IRConConnection`, config fields, API DTO fields
  - **Integrations/Cod + Source**: Stopwatch-based RTT measurement in both RCon implementations
  - **Application**: `LatencyService` (self-contained per-server service with event subscriptions and probe loop), DI registration, IW4MServer lifecycle wiring
  - **GSC**: `MonitorLatencyProbe` dvar polling loop in `_integration_shared.gsc`
  - **Webfront**: Latency fields populated in ServerInfo API responses

  ## API
  `GET /api/server` now includes per server:
  - `rconRoundTripMs` — EMA-smoothed RCON round-trip (all servers, ~24s to populate)
  - `logPipelineMs` — EMA-smoothed full log pipeline latency (GSC required, ~90s to populate)

  Both return `null` until 3 samples are collected.


Application/IW4MServer.cs
```diff
@@ -1554,6 +1556,10 @@ await this.SetDvarAsync("sv_sayname", CustomSayName,
             await _serverCache.InitializeAsync();
             _ = Task.Run(() => LogEvent.PollForChanges());
 
+            LatencyMetrics = new ServerLatencyMetrics(_appConfig.LatencyEmaAlpha);
```
RaidMax: If we are creating an new service instance for every server, I think it would make more sense to just proxy it through the getter. eg `LatencyMetrics => _latencyService.LatnecyMetrics` or something

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+{
+    private readonly ConcurrentDictionary<string, DateTime> _pendingProbes = new();
+
+    private CancellationTokenSource _cts;
+    private Task _probeLoopTask;
+    private bool _gscDetected;
+
+    private static readonly TimeSpan StaleProbeThreshold = TimeSpan.FromSeconds(30);
+
+    public async Task StartAsync()
```
RaidMax: This isn't doing any async work. Can me just make this void Start()?

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+            _probeLoopTask = Task.Run(() => ProbeLoopAsync(_cts.Token));
+        }
+    }
+
+    public async Task StopAsync()
+    {
+        IGameServerEventSubscriptions.ServerStatusReceived -= OnServerStatusReceived;
+        IGameEventSubscriptions.ScriptEventTriggered -= OnScriptEventTriggered;
+
+        if (_cts is not null)
```
RaidMax: Thinking it may make more sense to follow some of the existing patterns where we use the application manager cancellation token to trigger a stop instead of a call this is method and manual cancellation token handling.

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private async Task SendProbeAsync(CancellationToken token)
+    {
+        var probeId = Guid.NewGuid().ToString("N")[..8];
+        _pendingProbes[probeId] = DateTime.UtcNow;
+
+        try
+        {
+            await server.SetDvarAsync("sv_iw4madmin_probe", probeId, token);
```
RaidMax: Consider storing this string as a class level const

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+                logger.LogDebug(ex, "Failed to send latency probe");
+            }
+        }
+    }
+
+    private async Task DetectGscCompanionAsync()
+    {
+        try
+        {
+            var dvar = await server.GetDvarAsync("sv_iw4madmin_latencyprobe", "0");
```
RaidMax: Consider storing this string as a class level const

Application/Services/ServerLatencyMonitoringService.cs
```diff
@@ -0,0 +1,211 @@
...
+            _pendingProbes.TryRemove(probeId, out _);
+
+            using (LogContext.PushProperty("Server", server.Id))
+            {
+                logger.LogDebug(ex, "Failed to send latency probe");
+            }
+        }
+    }
+
+    private async Task DetectGscCompanionAsync()
```
RaidMax, line 189: Do we actually expect this to change regularly after a server is started? Ends up being 2x the queries when it seems like once the server is started it's unlikely the owner will be changing it live. Maybe check when the map starts?

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        return Task.CompletedTask;
+    }
+
+    private async Task ProbeLoopAsync(CancellationToken token)
+    {
+        // Wait for GSC companion to initialize — retry detection each interval
+        while (!token.IsCancellationRequested && !_gscDetected)
+        {
+            await Task.Delay(appConfig.LatencyProbeIntervalMs, token);
```
RaidMax: Seems like rather than a loop a timer approach is appropriate here. I think that would simplify this as well as the dangling background task.

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+                logger.LogDebug(ex, "Could not detect GSC latency probe companion");
+            }
+        }
+    }
+
+    private void CleanStalePendingProbes()
+    {
+        var cutoff = DateTime.UtcNow - StaleProbeThreshold;
+
+        foreach (var kvp in _pendingProbes.Where(kvp => kvp.Value < cutoff))
```
RaidMax: Is iterating on the dictionary kvps thread safe? Seems like it could throw a collection changed exception in the ienumerable iteration. Please verify.

Application/Services/LatencyService.cs
```diff
@@ -0,0 +1,211 @@
...
+                logger.LogDebug("Sent latency probe {ProbeId}", probeId);
+            }
+        }
+        catch (Exception ex)
+        {
+            _pendingProbes.TryRemove(probeId, out _);
+
+            using (LogContext.PushProperty("Server", server.Id))
+            {
+                logger.LogDebug(ex, "Failed to send latency probe");
```
RaidMax: Should this be logged as an error or warning?

GameFiles/GameInterface/_integration_shared.gsc
```diff
@@ -63,16 +63,36 @@ Setup()
     level notify( level.notifyTypes.sharedFunctionsInitialized );
     level waittill( level.notifyTypes.gameFunctionsInitialized );
 
-    scripts\_integration_base::_SetDvarIfUninitialized( level.commonKeys.autoBalance, 0 ); 
+    SetDvar( "sv_iw4madmin_latencyprobe", "1" );
```
RaidMax: We need to use the `scripts\_integration_base::_SetDvarIfUninitialized` instead of the direct `SetDvar` since `SetDvar` does not exist on all supported game interface games.

RaidMax: Also, the dvar name (and the next one) should probably be defined above in the commonKeys like the autobalance dvar

GameFiles/GameInterface/_integration_shared.gsc
```diff
@@ -63,16 +63,36 @@ Setup()
...
         return;
     }
-    
+
     thread OnPlayerConnect();
 }
 
+MonitorLatencyProbe()
+{
+    level endon( "game_ended" );
```
RaidMax: event name needs to be `level.eventTypes.gameEnd `

GameFiles/GameInterface/_integration_shared.gsc
```diff
@@ -63,16 +63,36 @@ Setup()
...
+
+    for ( ;; )
+    {
+        probeValue = GetDvar( "sv_iw4madmin_probe" );
```
RaidMax: Same comments here for the dvar string reference

SharedLibraryCore/Dtos/ServerInfo.cs
```diff
@@ -50,5 +50,7 @@ public double? LobbyZScore
             }
         }
         public Reference.Game Game { get; set; }
+        public double? RconRoundTripMs { get; set; }
+        public double? LogPipelineMs { get; set; }
```
RaidMax: nit: consider renaming `GameLogPipelineMs` for clarity

SharedLibraryCore/Events/Game/LatencyProbeScriptEvent.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace SharedLibraryCore.Events.Game;
+
+public class LatencyProbeScriptEvent : GameScriptEvent, IGameScriptEvent
+{
+    public string ProbeId { get; set; }
+
+    // Explicit interface implementation so it doesn't appear as a DeclaredOnly property
+    // (which would break positional argument parsing in ParseArguments)
+    string IGameScriptEvent.EventName => "LP";
```
RaidMax: You can keep it like this if you'd like, but just for information purposes the default parsing takes the text before the ScriptEvent class name (in this case `LatencyProbe` ) and uses that as the default log event name parsing. So if you use `LatencyProbe;` instead of `LP;` in the GSC side it will automatically pick it up and you don't need this property implementation.

RaidMax: Image

SharedLibraryCore/Helpers/ServerLatencyMetrics.cs
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            return Math.Max(0, log.Value - rtt.Value / 2.0);
+        }
+    }
+
+    /// <summary>
+    /// Whether the GSC latency probe companion is detected and active on this server.
+    /// </summary>
+    public bool LogProbeEnabled { get; set; }
```
RaidMax: Not sure this makes sense to store here? Wouldn't the ServerLatentcyMetrics instance on the LatencyService just be unitialized?

SharedLibraryCore/Server.cs
```diff
@@ -148,6 +148,7 @@ public int ClientNum
         public string WorkingDirectory { get; protected set; }
         public IRConConnection RemoteConnection { get; protected set; }
         public IRConParser RconParser { get; set; }
+        public ServerLatencyMetrics LatencyMetrics { get; set; } = new();
```
RaidMax: Per the other comments might make sense to make this nullable (proxied to the LatencyService member instance) and keep it null if the the LatencyService isn't initialized

Application/Services/ServerLatencyMonitoringService.cs
RaidMax, line 1: Do we anticipate measuring non server related latency here? Thinking it could be more descriptive to name this `ServerLatencyMonitoringService` or something similar

Application/IW4MServer.cs
Ayymoss: Done. Service now owns ServerLatencyMetrics instance, IW4MServer proxies it. Nullable when service not initialized.

Application/Services/LatencyService.cs
Ayymoss: Done. Changed to void Start(CancellationToken).

Application/Services/LatencyService.cs
Ayymoss: Done. Takes Manager.CancellationToken, cleanup via token.Register().

Application/Services/LatencyService.cs
Ayymoss: Done. Extracted to class-level const DvarProbe.

Application/Services/LatencyService.cs
Ayymoss: Done. Extracted to class-level const DvarLatencyProbe.

Application/Services/ServerLatencyMonitoringService.cs
Ayymoss: Good point. Moved to MatchStarted event subscription plus a one-shot initial detection timer for when IW4MAdmin connects mid-map.

Application/Services/LatencyService.cs
Ayymoss: Done. Replaced with System.Threading.Timer.

Application/Services/LatencyService.cs
Ayymoss: Good catch. Now takes a .ToArray() snapshot before iterating.

Application/Services/LatencyService.cs
Ayymoss: Changed to LogWarning.

GameFiles/GameInterface/_integration_shared.gsc
Ayymoss: Done. Using _SetDvarIfUninitialized for both dvars.

Ayymoss: Done. Added level.commonKeys.latencyProbe and level.commonKeys.probe, referenced throughout.

GameFiles/GameInterface/_integration_shared.gsc
Ayymoss: Done.

GameFiles/GameInterface/_integration_shared.gsc
Ayymoss: Done. All dvar references use commonKeys now.

SharedLibraryCore/Dtos/ServerInfo.cs
Ayymoss: Done. Renamed to GameLogPipelineMs everywhere.

SharedLibraryCore/Events/Game/LatencyProbeScriptEvent.cs
Ayymoss: Great tip. Switched GSC to use LatencyProbe; prefix and removed the explicit EventName implementation entirely.

Ayymoss: Thanks for the reference, this was helpful.

SharedLibraryCore/Helpers/ServerLatencyMetrics.cs
Ayymoss: Agreed. Removed LogProbeEnabled from ServerLatencyMetrics.

SharedLibraryCore/Server.cs
Ayymoss: Done. Nullable, proxied from the service instance. Null until service is initialized.

Application/Services/ServerLatencyMonitoringService.cs
Ayymoss: Done. Renamed to ServerLatencyMonitoringService.

RaidMax (approved):

PR #402, opened 2026-05-31
Gracefully handle plugins that target a newer IW4MAdmin API

Ayymoss:
## Problem

  When a plugin is recompiled against the latest SharedLibraryCore and published to the plugin store, older IW4MAdmin
  instances download it and try to load it. If the plugin actually uses API the older host doesn't have, the host throws
  — and because plugins are resolved in bulk via `IEnumerable<IPluginV2>` / `IEnumerable<IPlugin>`, a single bad plugin
  took the whole host down ("Failed to initialize IW4MAdmin").

  We want two things:
  1. **Don't break plugins that still work.** A plugin versioned against the latest API but only using members the older
  host *does* have should continue to load and run normally.
  2. **Soft-fail the rest.** A plugin that genuinely uses unavailable API should produce a single friendly notice
  ("update IW4MAdmin to load it") instead of crashing the host.

  A minimum-version attribute was deliberately rejected: it requires the author to know the correct minimum, which isn't
  mechanically inferable without a complete API-change lookup table we've never maintained. Instead this keys off the
  runtime exceptions the CLR already raises for missing types/members.

  ## How

  `SharedLibraryCore.Helpers.PluginApiCompatibility` centralises the logic (in SLC so both the host and the event
  dispatch can use it):
  - Classifies missing-API exceptions (`TypeLoadException`, `MissingMethod/Field/MemberException`,
  `FileNotFound/LoadException`), walking inner and `ReflectionTypeLoadException` loader exceptions.
  - `GetLoadableTypes` returns the types that *do* resolve and drops the ones that don't — so a plugin that doesn't
  actually touch the missing API loads normally.
  - Reports at most once per plugin (console + log), identified by assembly name.

  Missing-API failures are caught at every plugin execution boundary:
  - **Type discovery** (`PluginImporter`, eventable-type scan) — unresolvable types are dropped.
  - **Construction** — plugins are registered through a guarded factory; a plugin that loads its type fine but throws at
  construction (a ctor-body call, or a member typed with a missing type — neither is resolved at type load) is reported
  and replaced with an inert `UnavailablePlugin` placeholder, so the bulk DI resolve no longer fails. Covers
  `IPluginV2` and legacy `IPlugin`.
  - **Event dispatch** (`EventExtensions.RunHandler`), **`OnLoadAsync`**, **command/plugin event processing**
  (`IW4MServer`), and **`RegisterDependencies`**.

  Also fixes a latent bug surfaced during testing: a plugin loaded from the Plugins directory is also present in the
  AppDomain, so discovery listed it twice — registering and instantiating every such plugin twice. Discovery now dedupes
  by assembly.

  ## Coverage

  | Failure mode | Result |
  |---|---|
  | Missing base type / interface (eager at type load) | dropped at discovery + notice |
  | Missing type in field/property/param (lazy) | soft-fail at construction + notice |
  | Missing API in ctor body | soft-fail at construction + notice |
  | Missing API in handler/method body | soft-fail at dispatch + notice |

  Genuine plugin bugs (NRE, bad config) are unaffected — the missing-API catch is exception-filtered, so real errors
  still surface as errors rather than being swallowed.

  ## Testing

  Verified on an older host (SLC with the new API removed) against plugins compiled against the new API: type-load,
  construction (both member-signature and ctor-body), and handler-runtime cases each soft-fail with one notice and the
  host boots normally. Also validated against a real WIP plugin (ZombieStats) targeting newer API.

  ## Known limitation

  A plugin whose constructor subscribes to a static event *before* throwing on missing API leaves that subscription
  bound to a half-constructed instance. When it fires, missing-API calls soft-fail at dispatch; any other throw is
  logged and swallowed. This is log noise, not a crash, and a strict improvement over the previous fatal behaviour

### DIFF
--- a/Application/Main.cs
+++ b/Application/Main.cs
@@ -106,6 +106,8 @@ namespace IW4MAdmin.Application
 
             Console.CancelKeyPress += OnCancelKey;
             AppDomain.CurrentDomain.ProcessExit += OnProcessExit;
+            AppDomain.CurrentDomain.UnhandledException += OnUnhandledException;
+            TaskScheduler.UnobservedTaskException += OnUnobservedTaskException;
 
             Console.WriteLine("=====================================================");
             Console.WriteLine(" IW4MAdmin");
@@ -175,6 +177,40 @@ namespace IW4MAdmin.Application
         {
             Utilities.DefaultLogger.LogDebug("ProcessExit event received, performing synchronous shutdown");
             PerformShutdownAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Last-resort handler for an exception that escaped to a background/thread-pool thread —
+        /// typically a plugin <c>async void</c> handler or an unguarded fire-and-forget. The runtime
+        /// terminates the process after this returns (<see cref="UnhandledExceptionEventArgs.IsTerminating"/>
+        /// is set and there is no API to mark it handled), so we cannot recover. We can, however,
+        /// persist it to the IW4MAdmin log instead of losing it to stdout/stderr — which is gone the
+        /// moment a container is recreated. The Serilog file sink writes synchronously, so the entry
+        /// reaches disk before teardown.
+        /// </summary>
+        private static void OnUnhandledException(object sender, UnhandledExceptionEventArgs e)
+        {
+            var exception = e.ExceptionObject as Exception;
+            Utilities.DefaultLogger?.LogCritical(exception,
+                "Unhandled exception escaped to the runtime (terminating: {IsTerminating}). This is almost " +
+                "always a plugin throwing on a background thread, which IW4MAdmin cannot catch or recover from",
+                e.IsTerminating);
+
+            // belt-and-braces: the process is dying; make sure it also lands on stderr in case the
+            // logger was not yet initialised (very early startup)
+            Console.Error.WriteLine($"[FATAL] Unhandled exception: {exception}");
+        }
+
+        /// <summary>
+        /// Handler for a faulted <see cref="Task"/> that was never awaited/observed. Since .NET Core
+        /// these no longer terminate the process, so without this they vanish silently — usually a
+        /// plugin fire-and-forget bug. We log it and mark it observed so the behaviour stays explicit.
+        /// </summary>
+        private static void OnUnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
+        {
+            Utilities.DefaultLogger?.LogError(e.Exception,
+                "Unobserved task exception (likely a plugin fire-and-forget that faulted); marking observed");
+            e.SetObserved();
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


Application/Main.cs
```diff
@@ -474,7 +500,8 @@ private static IServiceCollection HandlePluginRegistration(ApplicationConfigurat
             // register any eventable types
             foreach (var assemblyType in typeof(Program).Assembly.GetTypes()
                          .Where(asmType => typeof(IRegisterEvent).IsAssignableFrom(asmType))
-                         .Union(plugins.SelectMany(asm => asm.Assembly.GetTypes())
+                         .Union(plugins.Select(pluginType => pluginType.Assembly).Distinct()
```
RaidMax, line 538: Does this double distinct make a meaningful difference?

SharedLibraryCore/Helpers/PluginApiCompatibility.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        return null;
+    }
+
+    /// <summary>
+    /// emits a user-facing notice (console + log) that a plugin could not be loaded/run because it
+    /// targets a newer IW4MAdmin API. Reports at most once per plugin for the life of the process.
+    /// </summary>
+    public static void NotifyNewerApiRequired(Assembly assembly, ILogger logger, string displayName = null)
```
RaidMax: Is there a way we could make this not static so we don't have to manually pass in the logger instance and use DI instead?

RaidMax (approved):

Ayymoss:
Your comments actioned. Also, a small additive change is included below:

Plugins that throw in an async void NullReferenceException on a thread-pool thread. It (1) never made it to the IW4MAdmin logs and (2) took the whole process down - and because it only went to stdout/stderr, the trace was lost the moment Docker recreated the container.

This adds two process-level handlers in Main.cs:

  - AppDomain.CurrentDomain.UnhandledException - logs the exception via Serilog (which writes synchronously, so it reaches disk before teardown). To be clear about the limit: an async void that escapes to a background thread is terminal in .NET — IsTerminating is set, and there's no API to mark it handled, so the process still exits. This can't prevent that crash; it just makes sure we have the trace instead of losing it. The plugin-side bug is fixed separately.
  - TaskScheduler.UnobservedTaskException - logs and calls SetObserved(). Faulted fire-and-forget Tasks no longer crash the process by default since .NET Core, so today they vanish silently; this surfaces them (usually a plugin bug) without changing runtime behaviour.

Both are purely additive - no effect on a healthy run or on working plugins. They only do anything when something has already escaped the normal guarded paths.

Application/Main.cs
```diff
@@ -177,6 +179,40 @@ private static void OnProcessExit(object sender, EventArgs e)
...
+        /// </summary>
+        private static void OnUnhandledException(object sender, UnhandledExceptionEventArgs e)
+        {
+            var exception = e.ExceptionObject as Exception;
+            Utilities.DefaultLogger?.LogCritical(exception,
+                "Unhandled exception escaped to the runtime (terminating: {IsTerminating}). This is almost " +
+                "always a plugin throwing on a background thread, which IW4MAdmin cannot catch or recover from",
+                e.IsTerminating);
+
+            // belt-and-braces: the process is dying; make sure it also lands on stderr in case the
```
RaidMax, line 199: FYI this is a very British idiom

PR #405, opened 2026-06-07
Plugin web resources: in-memory bundle loading + shared component kit

Ayymoss:
## Summary

After the webfront's MCP→Blazor migration, plugins can register routes and components dynamically (`AddAdditionalAssemblies`), but they **cannot ship their own web resources**. The static-asset manifest and the Tailwind purge are both compile-time, so a third-party or premium plugin's CSS/JS — and any Tailwind classes the host never compiled — 404 or render unstyled.

This PR adds the infrastructure for plugins to deliver web pages, in two layers.

### 1. Plugin web bundles

A plugin ships a single `.zip` (`manifest.json` + `lib/` + `wwwroot/` + optional `gsc/`). The host loads it **in memory only** — the DLL and web assets are never written to disk (premium anti-extraction; a decrypted DLL on disk could be copied into `Plugins/` to bypass licensing) — serves its assets at `/_content/{id}/…`, and extracts only sanctioned `gsc/` game scripts to disk.

- `IPluginAssetStore` + a dynamic-lookup `IFileProvider` bound into a dedicated `/_content` static-files branch, backed by a `ConcurrentDictionary` so late (remote-channel) registrations still serve.
- One process-wide `IPluginBundleLoader` cached by manifest id, so both load sites (`WebfrontCore.Program` for Razor/MVC, `Application.PluginImporter` for lifecycle) resolve the **same** `Assembly` — avoiding the duplicate-identity trap of `Assembly.Load(byte[])`. Everything stays in the default `AssemblyLoadContext` to preserve `IPluginV2`/DI type identity.
- MSBuild packaging targets shipped in the SharedLibraryCore package (`<IW4MAdminBundle>true</IW4MAdminBundle>`): build-is-bundle, tool-agnostic, with an optional removable Tailwind helper that emits into a `layer(plugins)` cascade layer so plugin utilities can never override host chrome.
- Remote-channel seam (`PluginType.Bundle`, `IRemoteAssemblyHandler.DecryptBundles`) for streaming encrypted bundles from the Master; inert until the Master side ships.
- `PluginGscExtractPath` config for the optional GSC extraction.

### 2. WebCommon shared component kit

A small Blazor RCL, published as its own package (`RaidMax.IW4MAdmin.WebCommon`), so plugin pages reuse the host's native UI instead of rolling their own:

- `SideContextMenu` **moved** out of WebfrontCore into WebCommon. Its `AppState` dependency is replaced by an optional `Localizer` delegate (host passes `AppState.Loc`), and the mobile drawer is reworked to be **CSS-only** (a hidden checkbox + `peer-checked:` + `<label>` toggles) so it works on statically-rendered plugin pages too — no Blazor interactivity required.
- `PluginPageShell` — the standard content + sticky right-pane page layout, so a plugin page matches a first-party stats/leaderboard page.
- The host Tailwind input gains `@source "../../../../WebCommon"` so the shared components' classes are compiled into the host stylesheet (loaded on every page); plugin pages render with host classes and never depend on their own Tailwind build for layout.
- The bundle targets mark `WebCommon.dll` (like `SharedLibraryCore.dll`/`Data.dll`) host-provided, so plugins never bundle it.
- CI: the SharedLibraryCore NuGet workflow now also builds/packs/publishes WebCommon, version-locked to SharedLibraryCore.

All first-party consumers of `SideContextMenu` (advanced stats, stats overview, server scoreboard, profile, privileged, home) were migrated; full solution builds clean.

## Reference implementation

A complete, working consumer of this infrastructure lives in the Credify plugin — converted to a web bundle, serving its own assets at `/_content/credify/`, and using `PluginPageShell` + `SideContextMenu` for its navigation: **https://github.com/Ayymoss/Credify/tree/feature/web-bundle**

It demonstrates the end-to-end path: `<IW4MAdminBundle>true</IW4MAdminBundle>` build → in-memory load → assets served → plugin pages rendering with host-themed chrome and a Tailwind class the host never compiled.

## Why this branch

Targeted at `feature/zombie-stats` as the integration branch for the upcoming **premium ZombieStats web pages**, which are the first first-party consumer of this infrastructure.

## Design notes

- The `/_content` branch is public (web assets are inherently public) and short-circuits before `MapStaticAssets`, so fingerprinting never touches plugin paths.
- MIME via a shared `FileExtensionContentTypeProvider`; `ServeUnknownFileTypes=false`.
- No hot-reload/unload; no custom ALC.
- v1 bundle plugins expose Razor pages (the validated path); MVC controllers on a `byte[]`-loaded assembly are untested and documented as a limitation.

## Testing

- Full solution builds clean (net10.0).
- In-memory provider spike verified: `GET /_content/{id}/…` returns the asset with correct content-type and nothing on disk.
- End-to-end proof: the Credify plugin (linked above) converted to a bundle, serving its own CSS and a Tailwind class the host never compiled.

## Out of scope / follow-ups

- Master (Flask) `.zip` content-type support — separate change on the Master repo.
- Premium ZombieStats web-page migration — follows once this is approved.
- Docs: `docs/plugin-web-bundles.md` + `PLUGIN_WEB_RESOURCES_DESIGN.md`.


Application/Misc/RemoteAssemblyHandler.cs
```diff
@@ -52,6 +52,14 @@ public IEnumerable<string> DecryptScripts(string[] encryptedScripts)
             return DecryptContent(encryptedScripts).Select(decryptedScript => Encoding.UTF8.GetString(decryptedScript));
         }
 
+        public IEnumerable<byte[]> DecryptBundles(string[] encryptedBundles)
+        {
+            // Return the decrypted zip plaintext to the caller. We can't zero it here (as
+            // DecryptAssemblies does) because the bundle loader needs to read the archive after this
+            // returns; the caller zeroes each buffer once the loader has copied out what it needs.
+            return DecryptContent(encryptedBundles);
```
RaidMax: Is this method needed? It seems like it's just a proxy for the existing one. Since it doesn't have any logic, could we just make the existing one public?

Application/Plugin/PluginImporter.cs
```diff
@@ -212,12 +217,136 @@ private IEnumerable<string> GetRemoteScripts()
...
+        /// payloads are decrypted to raw zip bytes, loaded in-memory through the same shared loader the
+        /// local-disk path uses (so the assembly identity matches the one Razor discovery resolves from
+        /// the cache), and their game scripts are extracted to disk. The decrypted zip plaintext is zeroed
+        /// once the loader has consumed it so a runnable copy of the premium DLL doesn't linger on the heap.
+        /// </summary>
+        private IEnumerable<Assembly> GetRemoteBundles()
+        {
+            try
+            {
+                _pluginSubscription ??= masterApi
```
RaidMax: I realized with the addition of this type we would have 3 checks on the `_pluginSubscription` potentially in parallel? as part of the enumeration? Let's make sure we don't need to do any thread synchronization as part of the null coalesce here.

Application/Plugin/PluginImporter.cs
```diff
@@ -212,12 +217,136 @@ private IEnumerable<string> GetRemoteScripts()
...
+                    return Enumerable.Empty<Assembly>();
+                }
+
+                var assemblies = new List<Assembly>();
+
+                foreach (var zipBytes in remoteAssemblyHandler.DecryptBundles(encryptedBundles))
+                {
+                    try
+                    {
+                        var bundle = bundleLoader.LoadFromZipBytes(zipBytes, "remote bundle");
```
RaidMax: What's this magic string "remote bundle" used for? If it's logically important we should make it a constant

Application/Plugin/PluginImporter.cs
```diff
@@ -212,12 +217,136 @@ private IEnumerable<string> GetRemoteScripts()
...
+                        if (bundle is null)
+                        {
+                            continue;
+                        }
+
+                        ExtractBundleGameScripts(bundle);
```
RaidMax, line 271: This kind of a has a code smell to it. We access a mutated property of passed parameter. I think we should return the extracted assembly from the method rather than accessing that mutated property.

Application/Plugin/PluginImporter.cs
```diff
@@ -212,12 +217,136 @@ private IEnumerable<string> GetRemoteScripts()
...
         }
+
+        /// <summary>
+        /// Loads plugin bundles streamed from the master (premium subscription channel). The encrypted
+        /// payloads are decrypted to raw zip bytes, loaded in-memory through the same shared loader the
+        /// local-disk path uses (so the assembly identity matches the one Razor discovery resolves from
+        /// the cache), and their game scripts are extracted to disk. The decrypted zip plaintext is zeroed
+        /// once the loader has consumed it so a runnable copy of the premium DLL doesn't linger on the heap.
+        /// </summary>
+        private IEnumerable<Assembly> GetRemoteBundles()
```
RaidMax: Inconsistent naming on this method. Local implementation `GetBundleAssemblies` suffixes "Assemblies", which I think better describes the method. it's omitted here

Application/Plugin/PluginImporter.cs
```diff
@@ -212,12 +217,136 @@ private IEnumerable<string> GetRemoteScripts()
...
+        /// <summary>
+        /// Loads plugin bundles (*.zip or pre-unpacked folders) from the plugins dir via the shared loader
+        /// (which caches by id, so the assembly identity matches the one already loaded for Razor discovery)
+        /// and extracts any bundled game scripts to disk.
+        /// </summary>
+        private IEnumerable<Assembly> GetBundleAssemblies(string pluginDir)
+        {
+            var assemblies = new List<Assembly>();
+
+            foreach (var source in BundleDiscovery.Enumerate(pluginDir))
```
RaidMax: This BundleDiscovery class seems potentially unneeded. For local plugin discovery (non bundle) we utilize the filesystem already in this class. So either we continue that practice and remove this new class, or we move all the file operations (for local discovery) into the BundleDiscovery class. I don't forsee any alternative plugin sourcing methods beside files system and network so I would recommend keeping file operations contained here.

Application/Main.cs
```diff
@@ -441,6 +441,8 @@ private static IServiceCollection HandlePluginRegistration(ApplicationConfigurat
                 .AddSingleton(appConfig)
                 .AddSingleton(masterApi)
                 .AddSingleton<IRemoteAssemblyHandler, RemoteAssemblyHandler>()
+                .AddSingleton<SharedLibraryCore.Interfaces.IPluginBundleLoader>(_ =>
```
RaidMax: Are these FQ namespaces necessary?

Application/Main.cs
```diff
@@ -641,6 +643,10 @@ private static void ConfigureServices(IServiceCollection serviceCollection)
                 .AddSingleton<IResourceQueryHelper<ChatSearchQuery, MessageResponse>, ChatResourceQueryHelper>()
                 .AddTransient<IParserPatternMatcher, ParserPatternMatcher>()
                 .AddSingleton<IRemoteAssemblyHandler, RemoteAssemblyHandler>()
+                .AddSingleton<SharedLibraryCore.Interfaces.IPluginAssetStore>(_ =>
+                    SharedLibraryCore.Plugins.InMemoryPluginAssetStore.Shared)
```
RaidMax: Are these FQ namespaces necessary?

SharedLibraryCore/Interfaces/IPluginAssetStore.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace SharedLibraryCore.Interfaces;
+
+/// <summary>
+/// In-memory store for plugin-supplied web assets (css/js/images) served under /_content/{pluginId}/.
+/// Assets are never written to disk; the backing store is mutated as plugins load (including remote
+/// plugins that register after the web host has started), so <see cref="FileProvider"/> must be a
+/// stable reference that performs live lookups.
+/// </summary>
+public interface IPluginAssetStore
```
RaidMax: nit: `Store` might be confused with the `https://store.raidmax.org/` plugin store. Perhaps rename to `IPluginAssetStorage`?

SharedLibraryCore/Plugins/BundleManifest.cs
```diff
@@ -0,0 +1,40 @@
...
+    public string WebRoot { get; set; } = "wwwroot";
+
+    /// <summary>Folder within the bundle holding game-side scripts extracted to disk. Defaults to "gsc".</summary>
+    public string GscRoot { get; set; } = "gsc";
+
+    /// <summary>Advisory in phase 1: plugins still self-register their navbar pages.</summary>
+    public List<BundlePage> Pages { get; set; } = new();
+}
+
+public class BundlePage
```
RaidMax: Might make sense to put in separate file

SharedLibraryCore/Plugins/InMemoryPluginAssetStore.cs
```diff
@@ -0,0 +1,89 @@
...
+/// persisted to disk. The exposed <see cref="FileProvider"/> performs live lookups so assets
+/// registered after the static-file middleware was built (remote plugins) are still served.
+/// </summary>
+public sealed class InMemoryPluginAssetStore : IPluginAssetStore
+{
+    /// <summary>
+    /// Process-wide instance. Exposed as a static so the WebfrontCore startup paths (which run before
+    /// the DI container is built) and the DI-resolved consumers share a single backing store.
+    /// </summary>
+    public static InMemoryPluginAssetStore Shared { get; } = new();
```
RaidMax: I don't like this but, I understand why

SharedLibraryCore/Plugins/InMemoryPluginAssetStore.cs
```diff
@@ -0,0 +1,89 @@
...
+    public bool TryGet(string requestPath, out byte[] content) =>
+        _assets.TryGetValue(Normalize(requestPath), out content!);
+
+    private static string Normalize(string path) =>
+        "/" + path.Replace('\\', '/').TrimStart('/');
+
+    /// <summary>
+    /// <see cref="IFileProvider"/> that resolves files from the owning store's live map.
+    /// </summary>
+    private sealed class InMemoryPluginFileProvider(InMemoryPluginAssetStore store) : IFileProvider
```
RaidMax: The following two classes should be in their own files

SharedLibraryCore/Plugins/InMemoryPluginAssetStore.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                return new InMemoryFileInfo(name, content);
+            }
+
+            return new NotFoundFileInfo(subpath);
+        }
+
+        public IDirectoryContents GetDirectoryContents(string subpath) => NotFoundDirectoryContents.Singleton;
+
+        public IChangeToken Watch(string filter) => NullChangeToken.Singleton;
```
RaidMax: Is this used? What is its purpose?

SharedLibraryCore/Plugins/InMemoryPluginAssetStore.cs
```diff
@@ -0,0 +1,89 @@
...
+    private sealed class InMemoryFileInfo(string name, byte[] content) : IFileInfo
+    {
+        public bool Exists => true;
+        public long Length => content.Length;
+        public string? PhysicalPath => null;
+        public string Name => name;
+        public DateTimeOffset LastModified => DateTimeOffset.UnixEpoch;
+        public bool IsDirectory => false;
+
+        public Stream CreateReadStream() => new MemoryStream(content, writable: false);
```
RaidMax: Let's ensure this is disposed properly by all consumers

SharedLibraryCore/Plugins/LoadedBundle.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace SharedLibraryCore.Plugins;
+
+/// <summary>
+/// Result of loading a plugin bundle: the loaded assembly plus its parsed manifest and resources.
+/// Web assets are held in memory (served via the asset store); GSC files are surfaced for optional
+/// extraction to the configured game-files folder.
+/// </summary>
+public sealed class LoadedBundle
+{
+    public required string Id { get; init; }
```
RaidMax: Doesn't the Id exist on the manifest? Do we need the Id set multiple places?

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    public LoadedBundle? LoadFromZipBytes(byte[] zipBytes, string sourceLabel)
+    {
+        try
+        {
+            using var memoryStream = new MemoryStream(zipBytes, writable: false);
+            using var archive = new ZipArchive(memoryStream, ZipArchiveMode.Read);
+
+            var manifest = ReadManifest(ReadEntryBytes(archive, "manifest.json"), sourceLabel);
```
RaidMax: Should extract magic string to constant

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+
+            lock (_loadLock)
+            {
+                if (_cache.TryGetValue(manifest.Id, out var cached))
+                {
+                    return cached;
+                }
+
+                var libAssemblies = ReadFolder(archive, LibFolder(manifest));
```
RaidMax, line 96: There's probably a way to do the following 4 operations in parallel. Not required, but if it's easy might be a little faster.

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+        {
+            throw new InvalidOperationException($"Failed to load plugin bundle '{sourceLabel}': {ex.Message}", ex);
+        }
+    }
+
+    public LoadedBundle? LoadFromDirectory(string directory)
+    {
+        try
+        {
+            var manifestPath = Path.Combine(directory, "manifest.json");
```
RaidMax: magic string

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+            if (!File.Exists(manifestPath))
+            {
+                throw new InvalidDataException("directory is missing manifest.json");
```
RaidMax: nit: "Directory"

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+            var manifest = ReadManifest(File.ReadAllBytes(manifestPath), directory);
+
+            lock (_loadLock)
+            {
+                if (_cache.TryGetValue(manifest.Id, out var cached))
+                {
+                    return cached;
+                }
+
+                var libAssemblies = ReadDirectory(directory, LibFolder(manifest));
```
RaidMax, line 131: Same comment about parallel here.

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+        catch (Exception ex)
+        {
+            throw new InvalidOperationException($"Failed to load plugin bundle directory '{directory}': {ex.Message}", ex);
+        }
+    }
+
+    // loads the entry assembly (and exposes its private deps) entirely in memory, for the encrypted/zip path
+    private static Assembly LoadEntryInMemory(BundleManifest manifest, IReadOnlyDictionary<string, byte[]> libAssemblies)
+    {
+        var entryName = Path.GetFileNameWithoutExtension(manifest.EntryAssembly.Replace('\\', '/').TrimStart('/'));
```
RaidMax: I've seen this .Replace and TrimStart logic a couple times now I think? Consider extracting to the Utilities class

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+            GscFiles = gscFiles
+        };
+
+        _cache[manifest.Id] = bundle;
+        return bundle;
+    }
+
+    private static string LibFolder(BundleManifest manifest)
+    {
+        var entry = manifest.EntryAssembly.Replace('\\', '/').TrimStart('/');
```
RaidMax: Extract per prev comment

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    private static BundleManifest ReadManifest(byte[]? manifestBytes, string sourceLabel)
+    {
+        if (manifestBytes is null)
+        {
+            throw new InvalidDataException("bundle is missing manifest.json");
+        }
+
+        var manifest = JsonSerializer.Deserialize<BundleManifest>(manifestBytes, JsonOptions)
```
RaidMax: JsonSerializer.Deserialize can throw an exception in addition to being null. Do we want to handle that with an informative message?

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+        {
+            throw new InvalidDataException($"bundle '{manifest.Id}' manifest has no entryAssembly");
+        }
+
+        return manifest;
+    }
+
+    private static byte[]? ReadEntryBytes(ZipArchive archive, string path)
+    {
+        var entry = archive.GetEntry(path.Replace('\\', '/').TrimStart('/'));
```
RaidMax: Extract reused replace/trimstart

SharedLibraryCore/Plugins/PluginBundleLoader.cs
```diff
@@ -0,0 +1,290 @@
...
+
+    private static IReadOnlyDictionary<string, byte[]> ReadFolder(ZipArchive archive, string folder)
+    {
+        var result = new Dictionary<string, byte[]>(StringComparer.OrdinalIgnoreCase);
+        if (string.IsNullOrWhiteSpace(folder))
+        {
+            return result;
+        }
+
+        var prefix = folder.Replace('\\', '/').Trim('/') + "/";
```
RaidMax: replace

PR #406, opened 2026-06-13
refactor(scriptplugins): port JS parsers and plugins to C# scripts

Ayymoss:
Convert all 20 RCon/event parsers and the VPNDetection and ServerBanner plugins from Jint JavaScript to Roslyn-compiled C# script plugins, using C#-first idioms (enums over magic numbers, typed config, HttpClient, Command classes, inlined EF queries).

Add IManager.AddOrReplace{RCon,Event}Parser helpers in SharedLibraryCore so parser scripts register idempotently. Legacy Jint infrastructure is retained (deprecated, not removed).

HGM is running this commit. All servers and players are registering. Commands are being processed.


Plugins/ScriptPlugins/ParserTeknoMW3.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        rconParser.Version = version;
+        rconParser.GameName = Server.Game.IW5;
+        rconParser.CanGenerateLogPath = false;
+
+        eventParser.Configuration.GameDirectory = "scripts";
+        eventParser.Version = version;
+        eventParser.GameName = Server.Game.IW5;
+
+        manager.AddOrReplaceRConParser(rconParser);
```
RaidMax: Parsers should not register themselves explicitly. The previous implementation handled it in a controlled fashion via the script plugin wrapper, which was not exposed to the plugin itself. Since we are moving them to .cs where we can register them with dependency injection, I think it makes sense to use DI on the initial load (during plugin discovery) to have them injected into the application manager. 

Dynamic reloads and adding new parsers after IW4MAdmin is running get a little trickier with this approach, but the previous implementation did not support that to begin with.

I don't think dynamic reload is needed for parsers outside of a development context, because changing the parser on the fly could lead to an inconsistent state. For example: if the initial dvar collection succeeded and then a change is made that breaks that , but fixes something else, you get in a weird state where it's actively connected, but on the next restart it will fail to connect.

Application/Plugin/CSharpScript/CsPluginServiceHost.cs
```diff
@@ -131,17 +132,36 @@ private async Task LoadPluginAsync(string pluginPath)
...
+            var pluginType = loadableTypes
+                .FirstOrDefault(t => t.GetInterface(nameof(IPluginV2)) != null);
+            var parserTypes = loadableTypes
+                .Where(t => t.GetInterface(nameof(IParserDefinition)) != null)
+                .ToList();
+
+            if (pluginType is null && parserTypes.Count == 0)
             {
                 var availableTypes = string.Join(", ", assembly.GetTypes().Select(t => t.FullName));
                 throw new InvalidOperationException(
```
RaidMax: nit: this isn't new, but do we really want to throw an exception if they just haven't implemented ipluginv2 or parser yet?

Application/Plugin/CSharpScript/CsPluginServiceHost.cs
```diff
@@ -203,6 +235,52 @@ private async Task UnloadPluginAsync(string pluginPath)
...
+    {
+        for (var i = list.Count - 1; i >= 0; i--)
+        {
+            if (string.Equals(keySelector(list[i]), name, StringComparison.Ordinal))
+            {
+                list.RemoveAt(i);
+            }
+        }
+    }
+
```
RaidMax: Both of these methods are janky.
I'm not a fan of having another class manipulating the list of Rcon/EventParser in the manager. 
(We solely do this as part of the original ScriptPlugin implementation for JS but that is because we can't DI properly with that) -- which I'm actually ok with removing that part entirely. I see no point of retaining JS parser support, since migration is easy and there shouldn't be many, if any,  3rd party JS parsers.

Also I'm confused by the seemingly contradicting language of "they have no lifecycle" which is then followed by code to Add/Replace and Remove which IS a lifecycle.

I think it would make more sense to have the IManager implementation inject the ICsPluginServiceHost singleton and expose the "LoadedParsers" like we do for the plugins currently.

Image

(Which by the way is unused)

Then the "AdditionalRconParsers" and "AdditionalEventParsers" can be refactored to project the parsers tracked by the CsPluginServiceHost instance

RaidMax (approved):

PR #407, opened 2026-06-15
feat(plugins): per-plugin data directories

Ayymoss:
## What
Each plugin gets an isolated `Plugins/<key>/` folder for its config, its own database, and any files it owns — instead of everything landing in the shared `Configuration/` and `Database/` folders. Three opt-in surfaces, all keyed off the plugin automatically:

```csharp
public static void RegisterDependencies(IServiceCollection services)
{
    services.AddConfiguration<MySettings>();                    // → Plugins/MyPlugin/MySettings.json
    services.AddConfiguration<MyTranslations>("translations");  // custom name → translations.json
    services.AddDatabase<MyDbContext>();                        // → Plugins/MyPlugin/MyDbContext.db (migrated + WAL)
}
```

For raw files, inject `IPluginDataDirectory<MyPlugin>`:

```csharp
string RootPath { get; }                        // Plugins/<key>/, created on first access
string GetPath(params string[] segments);       // resolve inside the folder; makes intermediate dirs; rejects ../ escape
Task<string> EnsureAsset(string resource, bool overwrite = false);  // copy an embedded resource out to the folder

// var cache = data.GetPath("cache", "tokens.bin");  → Plugins/MyPlugin/cache/tokens.bin (dirs created for you)
```

## Why
- `Configuration/` is a flat mix of host and every plugin's files (Credify alone ships ~30).
- A plugin with its own DB currently hand-rolls a path, a private-options `IDbContextFactory` (to dodge the non-generic `DbContextOptions` collision with the host context), migrate-on-load and WAL. `AddDatabase<T>` absorbs all of it.
- A plugin that just wants to drop a file on disk has no sanctioned home — it composes paths off `OperatingDirectory` by hand.

## How
- **Identity** comes from `typeof(T).Assembly` — no manual keys, no stack-walking. Same for compiled `.dll`, `.cs` scripts, `.zip` bundles, and remote in-memory plugins.
- **Routing by elimination:** host/framework assemblies keep the legacy flat `Configuration/`; everything else is a plugin.
- A plugin can register more than one config (settings, translations, …); the `name` argument disambiguates and may contain `/` to nest into a subfolder.
- **Migrations run before the plugin `Load` event;** a failed one disables only that plugin's DB, logged once.

## Compatibility
Additive and opt-in. A plugin that calls none of these — and the host's own configs — behave exactly as today. V2 configs auto-move out of `Configuration/` on first run; legacy V1 untouched.

Author guide: `docs/plugin-data-directories.md`. Validated end-to-end against Credify (config + its own SQLite DB) in its own repo — a net deletion of boilerplate there.

### DIFF
--- a/Application/ApplicationManager.cs
+++ b/Application/ApplicationManager.cs
@@ -29,6 +29,7 @@ using IW4MAdmin.Application.Configuration;
 using IW4MAdmin.Application.IO;
 using IW4MAdmin.Application.Migration;
 using IW4MAdmin.Application.Plugin.Script;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Serilog.Context;
@@ -344,8 +345,10 @@ namespace IW4MAdmin.Application
 
         /// <summary>
         /// Applies pending migrations for every plugin database registered via <c>AddDatabase</c>, before
-        /// plugins receive the <c>Load</c> event. Each migration runs independently and soft-fails: a bad
-        /// plugin migration disables only that database, never the host or other plugins.
+        /// plugins receive the <c>Load</c> event. The registration is pure data (a context accessor); the
+        /// host owns the setup here — apply migrations, then set WAL. Each database is set up independently
+        /// and soft-fails: a bad plugin migration disables only that database, never the host or other
+        /// plugins.
         /// </summary>
         private async Task MigratePluginDatabasesAsync(CancellationToken token)
         {
@@ -359,7 +362,11 @@ namespace IW4MAdmin.Application
             {
                 try
                 {
-                    await registration.MigrateAsync(token);
+                    await using var context = registration.CreateContext();
+                    await context.Database.MigrateAsync(token);
+                    // WAL lets readers proceed during writes; it is a persistent setting, so applying it
+                    // once after migration is enough. Plugin databases are SQLite by design.
+                    await context.Database.ExecuteSqlRawAsync("PRAGMA journal_mode=WAL;", token);
                     _logger.LogInformation("Migrated plugin database {Context} at {Path}",
                         registration.ContextName, registration.DatabasePath);
                 }
--- a/Application/ApplicationManager.cs
+++ b/Application/ApplicationManager.cs
@@ -5,6 +5,7 @@ using SharedLibraryCore;
 using SharedLibraryCore.Commands;
 using SharedLibraryCore.Configuration;
 using SharedLibraryCore.Configuration.Validation;
+using SharedLibraryCore.Database;
 using SharedLibraryCore.Database.Models;
 using SharedLibraryCore.Exceptions;
 using SharedLibraryCore.Helpers;
@@ -341,6 +342,36 @@ namespace IW4MAdmin.Application
             await server.ProcessUpdatesAsync(CancellationToken.None);
         }
 
+        /// <summary>
+        /// Applies pending migrations for every plugin database registered via <c>AddDatabase</c>, before
+        /// plugins receive the <c>Load</c> event. Each migration runs independently and soft-fails: a bad
+        /// plugin migration disables only that database, never the host or other plugins.
+        /// </summary>
+        private async Task MigratePluginDatabasesAsync(CancellationToken token)
+        {
+            var registrations = _serviceProvider.GetServices<PluginDatabaseRegistration>().ToList();
+            if (registrations.Count == 0)
+            {
+                return;
+            }
+
+            foreach (var registration in registrations)
+            {
+                try
+                {
+                    await registration.MigrateAsync(token);
+                    _logger.LogInformation("Migrated plugin database {Context} at {Path}",
+                        registration.ContextName, registration.DatabasePath);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex,
+                        "Could not migrate plugin database {Context} at {Path}; this plugin's database will be unavailable",
+                        registration.ContextName, registration.DatabasePath);
+                }
+            }
+        }
+
         public async Task Init()
         {
             IsRunning = true;
@@ -351,6 +382,7 @@ namespace IW4MAdmin.Application
             Console.WriteLine(_translationLookup["MANAGER_MIGRATION_START"]);
             await ContextSeed.Seed(_serviceProvider.GetRequiredService<IDatabaseContextFactory>(), _isRunningTokenSource.Token);
             await DatabaseHousekeeping.RemoveOldRatings(_serviceProvider.GetRequiredService<IDatabaseContextFactory>(), _isRunningTokenSource.Token);
+            await MigratePluginDatabasesAsync(_isRunningTokenSource.Token);
             _logger.LogInformation("Finished database migration sync");
             Console.WriteLine(_translationLookup["MANAGER_MIGRATION_END"]);
             #endregion
--- a/Application/IO/BaseConfigurationHandlerV2.cs
+++ b/Application/IO/BaseConfigurationHandlerV2.cs
@@ -8,6 +8,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using SharedLibraryCore;
+using SharedLibraryCore.Helpers;
 using SharedLibraryCore.Interfaces;
 
 namespace IW4MAdmin.Application.IO;
@@ -56,14 +57,10 @@ public class BaseConfigurationHandlerV2<TConfigurationType> : IConfigurationHand
             return defaultConfiguration;
         }
 
-        var cleanName = configurationName.Replace("\\", "").Replace("/", "");
+        Filename = ResolveConfigurationPath(configurationName);
+        WarnIfTypeNotOwnedByPlugin();
+        MigrateLegacyConfiguration(configurationName, Filename);
 
-        if (string.IsNullOrWhiteSpace(configurationName))
-        {
-            return defaultConfiguration;
-        }
-
-        Filename = Path.Join(Utilities.OperatingDirectory, "Configuration", $"{cleanName}.json");
         TConfigurationType readConfiguration = null;
 
         try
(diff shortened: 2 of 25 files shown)


SharedLibraryCore/Utilities.cs
```diff
@@ -1348,7 +1350,40 @@ public static IServiceCollection AddConfiguration<TConfigurationType>(
...
+
+        /// <summary>
+        /// Registers a plugin-owned SQLite database. The file lands in the plugin's own data folder
+        /// (<c>Plugins/&lt;plugin&gt;/&lt;name&gt;.db</c>, derived from <typeparamref name="TContext"/>'s
+        /// assembly); <paramref name="name"/> defaults to the context type name and may contain
+        /// forward-slash subfolders. Injects an isolated <see cref="IDbContextFactory{TContext}"/> and
+        /// records a migration that the host applies at startup (before the plugin's <c>Load</c>). A
+        /// plugin may register more than one database.
+        /// </summary>
+        public static IServiceCollection AddDatabase<TContext>(this IServiceCollection serviceCollection,
```
RaidMax, line 1363: Seems a bit limited to only support SQLite database. I could definitely see plugins benefiting from using the same provider as the main IW4MAdmin instance. Would it not make sense to use the existing iw4madmin configuration for the actual database engine, but use plugin-specific database names? For SQLite it would mostly be the same, but for other providers you could create something like `IW4MAdmin_<PluginName> ` database or something

SharedLibraryCore/Utilities.cs
```diff
@@ -1348,7 +1350,40 @@ public static IServiceCollection AddConfiguration<TConfigurationType>(
...
+            var factory = new PluginDbContextFactory<TContext>($"Data Source={databasePath}");
+
+            serviceCollection.AddSingleton<IDbContextFactory<TContext>>(factory);
+
+            serviceCollection.AddSingleton(new PluginDatabaseRegistration(
+                typeof(TContext).Name,
+                databasePath,
+                async token =>
+                {
+                    await using var context = factory.CreateDbContext();
```
RaidMax: This section seems janky and would be kind of misleading in a stack trace. Is there any way we have have the plugin loading context handle the migrations and "setup" instead of the dependency injection resolver? I think typically dependency resolution should be similar to constructors where we strive for atomicity and avoid async stuff.

SharedLibraryCore/Interfaces/IPluginDataDirectory.cs
```diff
@@ -0,0 +1,37 @@
...
+///
+/// The folder name is derived from the plugin's own assembly and the directory is created on first
+/// access; callers only ever use relative paths inside it. This mirrors PaperMC's
+/// <c>getDataFolder()</c>: the host owns "where", the plugin owns "what".
+/// </summary>
+/// <typeparam name="TPlugin">
+/// The plugin's own type (any type defined in the plugin assembly). Its assembly identifies the
+/// folder, so the same instance is shared by everything in that plugin.
+/// </typeparam>
+public interface IPluginDataDirectory<TPlugin>
```
RaidMax: I would say rename this interface to better reflect what it's doing. It's named like a DTO, but you have method that perform actions. Alternatively consider having the plugin context handle this. I'm not 100% sure this breakout is needed.

SharedLibraryCore/Interfaces/IPageList.cs
```diff
@@ -1,9 +1,27 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 
 namespace SharedLibraryCore.Interfaces
 {
     public interface IPageList
```
RaidMax, line 5: No action needed because of backwards compatibility for plugins, but with these changes something like `IPageListRegistry` would be more accurate

SharedLibraryCore/Helpers/PluginDirectoryResolver.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        return combined;
+    }
+
+    /// <summary>
+    /// Strips characters illegal in a file name and defends against relative-path tokens, collapsing
+    /// the result to a safe, stable folder name. Never returns empty.
+    /// </summary>
+    public static string Sanitize(string name)
```
RaidMax, line 126: I think we probably have this same type of logic elsewhere. We should move this to utilities and check for any overlap.

SharedLibraryCore/Helpers/PluginDirectoryResolver.cs
```diff
@@ -0,0 +1,139 @@
...
+/// This avoids depending on membership in any discovery set (script assemblies are not enumerated
+/// there).
+/// </summary>
+public static class PluginDirectoryResolver
+{
+    /// <summary>
+    /// Framework/host assemblies whose types are *not* plugins. Types defined here keep the legacy
+    /// flat configuration location. Anything outside this set is treated as a plugin.
+    /// </summary>
+    private static readonly HashSet<string> HostAssemblies = new(StringComparer.OrdinalIgnoreCase)
```
RaidMax, line 26: I believe we have this collection defined elsewhere

SharedLibraryCore/Database/PluginDbContextFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly DbContextOptions<TContext> _options;
+
+    public PluginDbContextFactory(string connectionString)
+    {
+        _options = new DbContextOptionsBuilder<TContext>()
+            .UseSqlite(connectionString)
+            .Options;
+    }
+
+    public TContext CreateDbContext() =>
```
RaidMax, line 30: I wonder if there's a way we can just include this as a new method on the existing `DatabaseContextFactory`?

Application/ApplicationManager.cs
```diff
@@ -341,6 +342,36 @@ await Task.Delay(Math.Max(1000, _appConfig.RConPollRate - delayFactor),
...
         }
 
+        /// <summary>
+        /// Applies pending migrations for every plugin database registered via <c>AddDatabase</c>, before
+        /// plugins receive the <c>Load</c> event. Each migration runs independently and soft-fails: a bad
+        /// plugin migration disables only that database, never the host or other plugins.
+        /// </summary>
+        private async Task MigratePluginDatabasesAsync(CancellationToken token)
+        {
+            var registrations = _serviceProvider.GetServices<PluginDatabaseRegistration>().ToList();
```
RaidMax, line 355: Back to my previous comment on the migration in the dependency resolution. This makes sense here, why are we doing the migration there as well? I feel like we should be able to have a single migration location

Application/Plugin/PluginImporter.cs
```diff
@@ -423,34 +423,53 @@ private static IEnumerable<BundleSource> EnumerateBundleSources(string pluginDir
...
         /// <summary>
-        /// Writes a bundle's GSC files to the configured game-files folder. GSC is the only bundle content
-        /// written to disk; when no path is configured, extraction is skipped.
+        /// Extracts a bundle's data files to the plugin's own sandbox folder (<c>Plugins/&lt;id&gt;/</c>):
+        /// generic resources to <c>Resources/</c>, and game scripts to <c>gsc/</c>. Game scripts are treated as
+        /// a sandboxed resource — the instance owner copies them onto their game server; nothing is delivered to
+        /// a shared location, and no operator config is required.
         /// </summary>
-        private void ExtractBundleGameScripts(LoadedBundle bundle)
+        private void ExtractBundleToSandbox(LoadedBundle bundle)
```
RaidMax, line 434: Hmm I thought we reading the static resources directly from the zip instead of persisting them to disk. This gets back to the issue of mixing static resources and user data in the same folder.

Application/Plugin/PluginDataStore.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// </summary>
+    private static Stream ResolveManifestResource(Assembly assembly, string name)
+    {
+        var direct = assembly.GetManifestResourceStream(name);
+        if (direct is not null)
+        {
+            return direct;
+        }
+
+        var normalized = name.Replace('/', '.').Replace('\\', '.');
```
RaidMax, line 92: Pretty sure we use this logic elsewhere, let's try to prevent duplication if possible

Application/Plugin/PluginDataDirectory.cs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public async Task<string> EnsureAsset(string embeddedResourceName, bool overwrite = false)
+    {
+        if (string.IsNullOrWhiteSpace(embeddedResourceName))
+        {
+            throw new ArgumentException("Resource name cannot be null or empty.", nameof(embeddedResourceName));
+        }
+
+        var target = GetPath(embeddedResourceName.Split('/', '\\'));
```
RaidMax: This set already exists as Utilities.DirectorySeparatorChars, however we could just simplify it to `Path.DirectorySeparatorChar`

Application/Plugin/PluginImporter.cs
```diff
@@ -423,34 +423,53 @@ private static IEnumerable<BundleSource> EnumerateBundleSources(string pluginDir
...
             }
+        }
 
-            foreach (var (relativePath, content) in bundle.GscFiles)
+        /// <summary>
+        /// Writes bundle files to disk under <paramref name="targetPath"/>, preserving sub-paths. When
+        /// <paramref name="skipUnchanged"/> is set, a file already present at the same byte length is left
+        /// alone — avoids rewriting large data files (e.g. GeoIP databases) on every load.
+        /// </summary>
+        private static void WriteBundleFiles(IReadOnlyDictionary<string, byte[]> files, string targetPath,
```
RaidMax, line 461: What the difference between this and `PluginDataDirectory.EnsureAsset`?

SharedLibraryCore/Helpers/PluginDirectoryResolver.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// forward-slash subfolder segments. Each segment is sanitized, <paramref name="extension"/> is
+    /// appended if absent, intermediate directories are created, and the result is constrained to the
+    /// root (a name that would escape throws).
+    /// </summary>
+    public static string CombineWithinRoot(string root, string name, string extension)
+    {
+        var rootFull = Path.GetFullPath(root);
+
+        var segments = (name ?? string.Empty)
+            .Split(['/', '\\'], StringSplitOptions.RemoveEmptyEntries)
```
RaidMax: `Path.DirectorySeparatorChar` or `Utilities.DirectorySeparatorChars`

Application/Plugin/PluginDataDirectory.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace IW4MAdmin.Application.Plugin;
+
+/// <summary>
+/// Default <see cref="IPluginDataDirectory{TPlugin}"/> implementation. Resolves the folder from the
+/// plugin's own assembly (<see cref="PluginDirectoryResolver"/>) and lazily creates it. Origin of the
+/// assembly is irrelevant — disk DLL, in-memory compiled <c>.cs</c> script, local/remote <c>.zip</c>
+/// bundle, or a binary streamed from the remote store all resolve by assembly name alone.
+/// </summary>
+public sealed class PluginDataDirectory<TPlugin> : IPluginDataDirectory<TPlugin>
```
RaidMax: Thoughts on combining this with `PluginDirectoryResolver`? At a quick glace there seems to be a good bit of overlap in responsibilities

Application/Misc/PageList.cs
```diff
@@ -18,9 +16,25 @@ class PageList : IPageList
...
+
         public PageList()
         {
             Pages = new Dictionary<string, string>();
+            PageIcons = new Dictionary<string, string>();
+        }
+
+        public void AddPage(string name, string location, string icon = null)
+        {
+            Pages[name] = location;
```
RaidMax, line 32: Would this be at risk for concurrency issues? I'm thinking it should probably be fine?

SharedLibraryCore/Utilities.cs
Ayymoss: Discussed on Discord. We'd try to support plugins writing to the same database, as 'Plugin' prefixed tables, but the complexity does scale for this problem.

For now, we'll just maintain SQLite-only support.

Application/IO/BaseConfigurationHandlerV2.cs
```diff
@@ -211,4 +208,91 @@ private void CopyUpdatedProperties(TConfigurationType newConfiguration)
...
+            "Define configuration types inside your own plugin to keep its data folder isolated.",
+            typeof(TConfigurationType).Name, assembly.GetName().Name, PluginDirectoryResolver.ResolveKey(assembly));
+    }
+
+    /// <summary>
+    /// One-time relocation of a plugin configuration from the legacy shared <c>Configuration/</c> folder
+    /// into the plugin's own folder. Runs only for plugin configs, only when the target is absent and the
+    /// legacy file is present — so it is idempotent and cannot double-apply across restarts.
+    /// </summary>
+    private void MigrateLegacyConfiguration(string configurationName, string targetPath)
```
RaidMax, line 270: Since this would be a one off execution, I don't think it makes sense to leave this in the BaseConfigurationHandler indefinitely. We already have a `ConfigurationMigration` class for this exact purpose.

Application/Plugin/PluginImporter.cs
Ayymoss: Spoke on Discord about this.

PR #412, opened 2026-08-23
Add native 7 Days to Die support

CoreyUK:
## Summary

Adds first-class 7 Days to Die support to IW4MAdmin without requiring an external protocol-conversion bridge.

IW4MAdmin communicates directly with the 7DTD Telnet console and reads the native `server-output.log`. Remote log files can be accessed through IW4MAdmin-GameLogServer.

## Features

- Native 7DTD Telnet/RCon integration
- Dedicated 7DTD parser
- First-class `D7D` game identification
- Support for multiple local and remote 7DTD servers
- Player connection and disconnection tracking
- Chat message parsing
- Webfront kick, temporary ban and permanent ban support
- Correct kick and ban reason handling
- Player level displayed as score
- Zombie kills and player deaths displayed separately
- Empty server messages filtered
- Standard webfront activity-history behaviour
- Support for local logs and IW4MAdmin-GameLogServer
- Optional Live Radar support using the native 7DTD Web Dashboard
- Dynamic map support for Navezgane, generated worlds and custom seeds

## Live Radar security

7DTD Live Radar has additional protections because it exposes precise player coordinates and explored terrain:

- Authentication required for 7DTD radar endpoints
- Per-IP request rate limiting
- Server-side player-data caching
- Bounded Telnet response sizes
- Bounded map-tile response sizes
- PNG-only map tile responses
- HTTP redirects disabled when dashboard credentials are attached
- Dashboard credentials redacted from HTTP logs
- Dashboard token values validated before use
- Dashboard URLs restricted to HTTP and HTTPS
- Dashboard secrets loaded from a file rather than stored in configuration

## Networking

7DTD Telnet and GameLogServer HTTP traffic are not encrypted by their respective protocols.

Remote services should therefore be restricted with firewall rules and accessed through private networking, a VPN or an authenticated encrypted tunnel. These services should not be exposed directly to the public Internet.

## Configuration

Each server entry can define its own:

- Telnet address, port and password
- Local or remote `server-output.log` path
- GameLogServer endpoint
- Live Radar dashboard endpoint
- Dashboard token name and secret-file path

Configuration instructions and an example are included in:

`Integrations/SevenDaysToDie/README.md`

## Validation

- Full .NET 10 solution build completed successfully
- Webfront build completed successfully
- Plugin build completed successfully
- Development Docker image built successfully
- Player join and disconnect handling tested
- Chat parsing tested
- Kick and ban reason handling tested
- Player level, zombie kills and deaths tested
- Empty-server behaviour tested
- Live Radar metadata, player data and map tiles tested
- Blank server-message filtering tested

GitHub Actions:

https://github.com/CoreyUK/IW4M-Admin/actions/runs/32641566962


Plugins/LiveRadar/Controllers/RadarController.cs
```diff
@@ -51,10 +73,172 @@ public IActionResult Data(string serverId = null)
...
+            }
+
+            if (zoom is < 0 or > 4 || tileX is < -128 or > 128 || tileY is < -128 or > 128)
+            {
+                return BadRequest();
+            }
+
+            try
+            {
+                var dashboardUrl = GetConfigurationValue<Uri>(server, "LiveRadarUrl") ??
```
RaidMax: I don't love relying on external services for the map data image, but presuming this is an external system run by the hoster?

Plugins/LiveRadar/Controllers/RadarController.cs
```diff
@@ -51,10 +73,172 @@ public IActionResult Data(string serverId = null)
...
+                if (!string.Equals(response.Content.Headers.ContentType?.MediaType, "image/png",
+                        StringComparison.OrdinalIgnoreCase) || response.Content.Headers.ContentLength is > MaxTileBytes)
+                {
+                    return StatusCode(StatusCodes.Status502BadGateway);
+                }
+
+                var content = await ReadBoundedContentAsync(response.Content, MaxTileBytes,
+                    HttpContext.RequestAborted);
+                Response.Headers.CacheControl = "private, max-age=86400, stale-while-revalidate=604800";
+                Response.Headers["X-Content-Type-Options"] = "nosniff";
```
RaidMax: Most of the non request related logic should be moved into a service class instead of directly in the controller. eg file operations, http(s) requests, configuration lookups,

WebfrontCore/Core/Services/AppState.cs
```diff
@@ -86,6 +86,11 @@ public void SetUser(ClientInfo user)
 
     public string Loc(string key)
     {
+        if (key == "GAME_D7D")
```
RaidMax: Remove this. We can add it to the localization sheet.

WebfrontCore/Core/Services/WebfrontDataService.cs
```diff
@@ -168,8 +168,8 @@ await _serverDataViewer.ClientHistoryAsync(
                             Online = true,
                             LastConnection = p.client.LastConnection,
                             Score = p.client.Score,
-                            Kills = p.stats?.MatchData?.Kills ?? 0,
-                            Deaths = p.stats?.MatchData?.Deaths ?? 0,
+                            Kills = GetRConKills(p.client) ?? p.stats?.MatchData?.Kills ?? 0,
```
RaidMax: If this data is available through the game engine then we should reusing the existing kill/death eventing to track it through the existing stats behavior. I'm guessing it's just available through "rcon" instead of discrete event logs? If so we should check for changes to the kills/deaths within status update and trigger the appropriate kill/death events instead of creating this pseudo-tracking via "AdditionalProperties"

WebfrontCore/Core/Services/WebfrontDataService.cs
```diff
@@ -1819,6 +1819,25 @@ public async Task DisableTwoFactorAsync(int? clientId = null)
         await _clientService.Update(client);
     }
 
+    private static int? GetRConKills(EFClient client) =>
```
RaidMax: Per previous comments these methods should be able to removed

WebfrontCore/Components/Features/Servers/Components/ServerCard.razor
```diff
@@ -13,7 +13,7 @@
             <!-- 1. Game Tag -->
             <span
                 class="px-1.5 py-0.5 rounded bg-surface text-[10px] font-mono text-muted border border-line uppercase tracking-widest shrink-0 mr-3">
-                @Model.Game
+                @(Model.Game == Data.Models.Reference.Game.D7D ? "7DTD" : Model.Game)
```
RaidMax: Remove this. 7DTD should be added to the Game enum already,

SharedLibraryCore/Configuration/ServerConfiguration.cs
```diff
@@ -48,6 +48,15 @@ public class ServerConfiguration : IBaseConfiguration
         [ConfigurationOptional]
         public string CustomHostname { get; set; }
 
+        [ConfigurationOptional]
+        public Uri LiveRadarUrl { get; set; }
```
RaidMax: I think we should make these options a bit more generic. How about `ExternalWebUrl`, `ExternalWebTokenName` and `ExternalWebTokenSecretFile`
Can you also explain what the "SecretFile" is?
It may make sense to move this into a sub configuration eg

```csharp
public class ExternalWeb 
{
   public string Url {get;set;}
   public  string TokenName {get;set;}
   public string TokenFile {get;set;}
}
```

Plugins/ScriptPlugins/Parser7DaysToDie.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        rcon.CommandPrefixes.RConGetInfo = null;
+        rcon.CommandPrefixes.Kick = "kick {0} \"{1}\"";
+        rcon.CommandPrefixes.Ban = rcon.CommandPrefixes.Kick;
+        rcon.CommandPrefixes.TempBan = rcon.CommandPrefixes.Kick;
+        rcon.CommandPrefixes.Say = "say \"{0}\"";
+        rcon.CommandPrefixes.Tell = "tell {0} \"{1}\"";
+        rcon.NoticeLineSeparator = " ";
+
+        rcon.ColorCodeMapping.Clear();
```
RaidMax, line 72: I think you should be able to create a new instance with values in the initializer rather than clearing and loop adding

Plugins/ScriptPlugins/Parser7DaysToDie.cs
```diff
@@ -0,0 +1,97 @@
...
+        events.Say.Pattern =
+            @"^.*Chat \(from '(?:Steam_)?(\d+)', entity id '(\d+)', to '[^']+'\): '([^']*)': (.*)$";
+        events.Say.AddMapping(ParserRegex.GroupType.OriginNetworkId, 1);
+        events.Say.AddMapping(ParserRegex.GroupType.OriginClientNumber, 2);
+        events.Say.AddMapping(ParserRegex.GroupType.OriginName, 3);
+        events.Say.AddMapping(ParserRegex.GroupType.Message, 4);
+
+        eventParser.Version = version;
+        eventParser.GameName = Server.Game.D7D;
+        eventParser.URLProtocolFormat = "steam://connect/{{ip}}:26900";
```
RaidMax: Does the game _only_ support port 26900? If not this should be {{port}} instead of hardcoded

Plugins/ScriptPlugins/Parser7DaysToDie.cs
RaidMax: Or actually since it's just mapping to empty string I think all you need to do is just clear it and leave it.

Plugins/ScriptPlugins/Parser7DaysToDie.cs
```diff
@@ -0,0 +1,97 @@
...
+        rcon.Status.AddMapping(ParserRegex.GroupType.RConKills, 3);
+        rcon.Status.AddMapping(ParserRegex.GroupType.RConDeaths, 4);
+        rcon.Status.AddMapping(ParserRegex.GroupType.RConPing, 5);
+        rcon.Status.AddMapping(ParserRegex.GroupType.RConNetworkId, 6);
+        rcon.Status.AddMapping(ParserRegex.GroupType.RConName, 7);
+        rcon.Status.AddMapping(ParserRegex.GroupType.RConIpAddress, 8);
+
+        rcon.DefaultDvarValues.Add("version", version);
+        rcon.DefaultDvarValues.Add("sv_running", "1");
+        rcon.DefaultDvarValues.Add("sv_hostname", "7 Days to Die Server");
```
RaidMax: For these default items, I presume they are overridden once the status starts getting parsed?

Plugins/LiveRadar/Plugin.cs
```diff
@@ -131,9 +139,10 @@ private Task OnMonitoringStarted(MonitorStartEvent monitorEvent, CancellationTok
         {
             // if it's an IW4 game, with custom callbacks, we want to 
             // enable the live radar page
-            var shouldRegisterPage = monitorEvent.Server.GameCode != Reference.Game.IW4 ||
-                                     !monitorEvent.Server.IsLegacyGameIntegrationEnabled ||
-                                     _addedPage;
+            var supportsIw4 = monitorEvent.Server.GameCode == Reference.Game.IW4 &&
+                              monitorEvent.Server.IsLegacyGameIntegrationEnabled;
+            var supportsSevenDays = (int)monitorEvent.Server.GameCode == 15;
```
RaidMax: We should try to use the enum name rather than casting it to int.

Plugins/LiveRadar/Controllers/RadarController.cs
```diff
@@ -25,11 +38,19 @@ public IActionResult Map(string serverId = null)
                 return NotFound();
             }
 
-            var map = config.Maps.FirstOrDefault(map => map.Name == server.CurrentMap.Name);
+            if ((int)server.GameCode == 15)
```
RaidMax: Use named enum not int value

Plugins/LiveRadar/Components/LiveRadar.razor.cs
```diff
@@ -49,7 +54,16 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
             var mapUrl = $"/Radar/{ServerId}/Map";
             try
             {
-                await JS.InvokeVoidAsync("initLiveRadar", radarUrl, mapUrl);
+                if (IsSevenDaysServer)
```
RaidMax: I don't love this. Wondering if there's a way to move the 7dtd logic into the existing liveradar script, but handle the per-game logic there?

Plugins/LiveRadar/Components/LiveRadar.razor.cs
```diff
@@ -17,10 +17,14 @@ public partial class LiveRadar
...
 
+    private bool IsSevenDaysServer =>
+        Servers.FirstOrDefault(server => server.ToString() == ServerId) is { } selectedServer &&
+        (int)selectedServer.GameName == 15;
+
     protected override void OnInitialized()
     {
         Servers = Manager.GetServers()
-            .Where(server => server.GameName == Server.Game.IW4)
+            .Where(server => server.GameName == Server.Game.IW4 || (int)server.GameName == 15)
```
RaidMax: Used named enum.

Plugins/LiveRadar/Components/LiveRadar.razor.cs
```diff
@@ -17,10 +17,14 @@ public partial class LiveRadar
     private string SelectedServerName =>
         Servers.FirstOrDefault(s => s.ToString() == ServerId)?.Hostname ?? "Unknown Server";
 
+    private bool IsSevenDaysServer =>
+        Servers.FirstOrDefault(server => server.ToString() == ServerId) is { } selectedServer &&
+        (int)selectedServer.GameName == 15;
```
RaidMax: Use named enum

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
```diff
@@ -0,0 +1,512 @@
...
+            health = player.Health,
+            isAlive = player.Health > 0,
+            id = FormattableString.Invariant(
+                $"{player.EntityId}:{player.PositionX:F1}:{player.PositionZ:F1}:{player.RotationY:F1}")
+        });
+
+        _radarDataJson = JsonSerializer.Serialize(payload, RadarJsonOptions);
+        _radarDataExpiresAt = DateTime.UtcNow + RadarDataLifetime;
+        return _radarDataJson;
+    }
```
RaidMax: Live radar stuff should not be part of the RCON connection logic. The RCON connection should only be used for sending and receiving purpose-agnostic  data.

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
```diff
@@ -0,0 +1,512 @@
...
+    private readonly ILogger<SevenDaysToDieRConConnection> _logger;
+    private readonly SemaphoreSlim _queryLock = new(1, 1);
+    private readonly Dictionary<int, SevenDaysToDiePlayer> _playersBySlot = new();
+
+    private TcpClient _client;
+    private NetworkStream _stream;
+    private DateTime _metadataExpiresAt;
+    private DateTime _radarDataExpiresAt;
+    private string _radarDataJson = "[]";
+    private string _hostname = "7 Days to Die Server";
```
RaidMax, line 38: Why are all these thing hard coded here? Shouldn't the defaults in the parser take care of this?

Integrations/SevenDaysToDie/SevenDaysToDiePlayer.cs
```diff
@@ -0,0 +1,19 @@
+namespace Integrations.SevenDaysToDie;
+
+public sealed record SevenDaysToDiePlayer(
```
RaidMax: I think this should inherit from EFClient as a lot of the items are shared between. The parser should figure out how to map the items that already exist on EFClient.

Integrations/SevenDaysToDie/SevenDaysToDiePlayer.cs
```diff
@@ -0,0 +1,19 @@
...
+    int Slot,
+    int EntityId,
+    long NetworkId,
+    string Name,
+    string Address,
+    int Ping,
+    int Level,
+    int ZombieKills,
+    int Deaths,
+    double PositionX,
```
RaidMax: Should be able to use Vector3 for position and rotation instead of individual fields.

Application/IW4MServer.cs
```diff
@@ -956,6 +956,13 @@ private async Task OnClientUpdate(EFClient origin)
             client.Ping = origin.Ping;
             client.Score = origin.Score;
 
+            if (origin.GetAdditionalProperty<bool>("RConStatsAvailable"))
```
RaidMax: Per other comments this likely should be able to go away if we convert the rconkill/death change tracking to kill/death events.

If for some reason that is not doable it would make more sense to store a single object with all the fields in the AdditionalProperties instead of several different ones.

Application/IW4MServer.cs
```diff
@@ -1833,6 +1832,21 @@ await Manager.GetPenaltyService().RemoveActivePenalties(targetClient.AliasLink.A
             });
         }
 
+        private string BuildPenaltyKickCommand(EFClient activeClient, EFPenalty currentPenalty,
+            EFPenalty previousPenalty = null)
+        {
+            var commandServer = activeClient.CurrentServer;
+            var commandConfig = commandServer.RconParser.Configuration;
+            var penalty = previousPenalty ?? currentPenalty;
+            var notice = commandServer.GameName == Game.D7D
```
RaidMax: Can you explain why this is needed? Is there not a way to handle whatever constraint this is elsewhere?

Application/RConParsers/BaseRConParser.cs
```diff
@@ -335,6 +335,19 @@ private List<EFClient> ClientsFromStatus(string[] Status)
...
 
+                    if (Configuration.Status.GroupMapping.TryGetValue(ParserRegex.GroupType.RConKills,
+                            out var killsGroupIndex) &&
+                        Configuration.Status.GroupMapping.TryGetValue(ParserRegex.GroupType.RConDeaths,
+                            out var deathsGroupIndex) &&
+                        match.Values.Length > Math.Max(killsGroupIndex, deathsGroupIndex) &&
+                        int.TryParse(match.Values[killsGroupIndex], out var kills) &&
+                        int.TryParse(match.Values[deathsGroupIndex], out var deaths))
+                    {
+                        client.SetAdditionalProperty("RConStatsAvailable", true);
```
RaidMax: Similar comment to previous, I think this should be able to be converted to events.

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
```diff
@@ -0,0 +1,512 @@
...
+            }
+
+            var message = CleanConsoleText(Unquote(playerCommand.Groups["message"].Value.Trim()));
+            if (verb == "tell" && string.IsNullOrWhiteSpace(message))
+            {
+                _logger.LogDebug("Ignoring empty 7 Days to Die private message for slot {Slot}", slot);
+                return string.Empty;
+            }
+
+            return verb == "tell"
```
RaidMax: nit: there might be better way to handle these commands instead of a long line of ifs?

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
```diff
@@ -0,0 +1,512 @@
...
+    private static readonly TimeSpan MetadataLifetime = TimeSpan.FromMinutes(1);
+    private static readonly TimeSpan RadarDataLifetime = TimeSpan.FromSeconds(1);
+    private const int MaxResponseBytes = 4 * 1024 * 1024;
+    private static readonly JsonSerializerOptions RadarJsonOptions = new(JsonSerializerDefaults.Web);
+
+    private readonly IPEndPoint _endpoint;
+    private readonly string _password;
+    private readonly ILogger<SevenDaysToDieRConConnection> _logger;
+    private readonly SemaphoreSlim _queryLock = new(1, 1);
+    private readonly Dictionary<int, SevenDaysToDiePlayer> _playersBySlot = new();
```
RaidMax: I'm not sure if we should be storing previous player stats here? Does the "rcon" status return a comprehensive list or just changes? If the former I don't think we need this. (It looks like it's just used for the kick command which if we properly propagate the Slot # during parsing we should be able to send it back in the kick payload as well.

RaidMax (changes requested):

Plugins/LiveRadar/Controllers/RadarController.cs
CoreyUK: Yep, it is the official 7DTD Web Dashboard run alongside the game server by the hoster, rather than a third-party map service. I've clarified that in the integration README.

Plugins/LiveRadar/Controllers/RadarController.cs
CoreyUK: Moved the file access, HTTP requests, dashboard authentication, caching, metadata/player parsing, and tile validation into a dedicated SevenDaysToDieRadarService. The controller now just handles routing, auth, validation, and HTTP responses.

WebfrontCore/Core/Services/AppState.cs
CoreyUK: Removed. The game name can be added through the localization sheet.

WebfrontCore/Core/Services/WebfrontDataService.cs
CoreyUK: Moving to a new method in reading the server log for such events,

WebfrontCore/Core/Services/WebfrontDataService.cs
CoreyUK: I've reduced the fallback to one RConStatusStats object. I left the small accessors in place until we agree on the event shape, since generating normal PvP kill events for zombie/world activity could affect rating calculations.

WebfrontCore/Components/Features/Servers/Components/ServerCard.razor
CoreyUK: Removed the special-case label; it now uses the game enum like the other titles.

SharedLibraryCore/Configuration/ServerConfiguration.cs
CoreyUK: Changed this to a nested ExternalWeb configuration with Url, TokenName, and TokenFile. TokenFile is the local path to a file containing the dashboard token secret, so the secret doesn't need to live in IW4MAdminSettings.json. The README now explains it and includes the new shape.

Plugins/ScriptPlugins/Parser7DaysToDie.cs
CoreyUK: Simplified this to just clear the mapping. 7DTD doesn't use CoD color codes, so there is no need to add empty values back into it.

Plugins/ScriptPlugins/Parser7DaysToDie.cs
CoreyUK: Changed to steam://connect/{{ip}}:{{port}}.

Plugins/ScriptPlugins/Parser7DaysToDie.cs
CoreyUK: Yes. The connection now seeds its initial values from the parser defaults, then the supported values are replaced with live game preferences during status polling.

Plugins/LiveRadar/Plugin.cs
CoreyUK: Changed to Reference.Game.D7D.

Plugins/LiveRadar/Controllers/RadarController.cs
CoreyUK: Changed to the named D7D enum.

Plugins/LiveRadar/Components/LiveRadar.razor.cs
CoreyUK: Merged the 7DTD renderer into the existing liveradar.js entry point. It now selects the renderer from the map provider metadata, so the component no longer imports a separate game-specific script.

Plugins/LiveRadar/Components/LiveRadar.razor.cs
CoreyUK: Changed to the named D7D enum.

Plugins/LiveRadar/Components/LiveRadar.razor.cs
CoreyUK: Changed to the named D7D enum.

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
CoreyUK: Agreed. All radar metadata, player DTO, caching, and tile logic has been removed from the RCON connection and moved into SevenDaysToDieRadarService. The connection now only translates and transports purpose-agnostic commands and responses.

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
CoreyUK: The initial values now come from the parser's DefaultDvarValues in SetConfiguration. Live preferences replace the supported values once polling starts.

Integrations/SevenDaysToDie/SevenDaysToDiePlayer.cs
CoreyUK: SevenDaysToDiePlayer now inherits EFClient, and the parser maps the shared name, address, network ID, client number, ping, and score fields there.

Integrations/SevenDaysToDie/SevenDaysToDiePlayer.cs
CoreyUK: Position and rotation now use Vector3.

Application/IW4MServer.cs
CoreyUK: Removed the 7DTD penalty helper. For the RCON counters, I've changed the fallback to a single RConStatusStats object while the event-model question is open.

Application/IW4MServer.cs
CoreyUK: This was only there to avoid sending CoD-formatted penalty text to 7DTD. I've removed the game-specific helper and restored the normal penalty formatting path; the parser is responsible for its formatting constraints.

Application/RConParsers/BaseRConParser.cs
CoreyUK: This is the same event-shape issue as above: listplayers only exposes cumulative counters, and there isn't a real player victim for zombie kills or necessarily a player attacker for deaths. Switching to using the log file for such events

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
CoreyUK: Replaced the long if chain with a switch and grouped player-command handling.

Integrations/SevenDaysToDie/SevenDaysToDieRConConnection.cs
CoreyUK: It is a comprehensive list. The cache has been removed, and the entity ID is now propagated through the parser's AdditionalGroup/ConnectionClientId path so tell and kick can send it straight back to 7DTD.

WebfrontCore/Core/Services/WebfrontDataService.cs
RaidMax: We should use the zombie/world event like the existing code for zombie client tracking

Application/Localization/Configure.cs
```diff
@@ -15,6 +15,12 @@ namespace IW4MAdmin.Application.Localization
 {
     public static class Configure
     {
+        private static readonly IReadOnlyDictionary<string, string> BuiltInFallbacks =
+            new Dictionary<string, string>
+            {
+                ["GAME_D7D"] = "7 Days to Die"
```
RaidMax: Don't need any of this. I've added to the localization sheet. Does not need a fallback

Integrations/SevenDaysToDie/SevenDaysToDiePlayer.cs
```diff
@@ -0,0 +1,14 @@
+using Data.Models;
+using SharedLibraryCore.Database.Models;
+
+namespace Integrations.SevenDaysToDie;
+
+public sealed class SevenDaysToDiePlayer : EFClient
+{
+    public int EntityId { get; init; }
```
RaidMax: Shouldn't `EntityId` be encapsulated by `ClientId` on `EFClient`?

Integrations/SevenDaysToDie/SevenDaysToDiePlayer.cs
```diff
@@ -0,0 +1,14 @@
...
+    public int ZombieKills { get; init; }
+    public int PlayerDeaths { get; init; }
```
RaidMax: I think these should be able to be tracked by the existing stats kills/deaths

Integrations/SevenDaysToDie/SevenDaysToDiePlayerParser.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Text.RegularExpressions;
+using Data.Models;
+using SharedLibraryCore;
+using SharedLibraryCore.Database.Models;
+
+namespace Integrations.SevenDaysToDie;
+
+public static partial class SevenDaysToDiePlayerParser
+{
+    public static IReadOnlyList<SevenDaysToDiePlayer> Parse(string response)
```
RaidMax: This is fine for now but in the future we may want to consider seeing what it would look like to abstract out the RconParser and EventParser  registration into the actual game parser plugin itself so that we can completely override the existing `BaseRconParser` implementation for games that are completely different than idtech @Ayymoss

Plugins/LiveRadar/Controllers/RadarController.cs
```diff
@@ -1,60 +1,123 @@
...
+    [EnableRateLimiting("liveRadar")]
+    public async Task<IActionResult> Map(string serverId = null)
     {
-        private readonly IManager _manager = manager;
+        var server = FindServer(serverId);
+        if (server is null) return NotFound();
+        if (server.GameCode == Reference.Game.D7D)
+        {
+            if (!Authorized) return Unauthorized();
+            return await ExecuteSevenDaysRequest(server, () => sevenDaysRadar.GetMapAsync(server,
```
RaidMax: This request should still be moved into the `SevenDaysToDieRadarService`. My previous comments about requests can remain here was specifically about handling checks on the _incoming_ request, not any outgoing requests for data

Plugins/LiveRadar/Controllers/RadarController.cs
```diff
@@ -1,60 +1,123 @@
...
+    }
 
-            var radarInfo = server.GetClientsAsList()
-                .Select(client => client.GetAdditionalProperty<RadarDto>("LiveRadar")).ToList();
+    private Server FindServer(string serverId) => serverId is null
+        ? _manager.GetServers().FirstOrDefault()
+        : _manager.GetServers().FirstOrDefault(server => server.ToString() == serverId);
 
-            return Json(radarInfo);
+    private async Task<IActionResult> ExecuteSevenDaysRequest(Server server, Func<Task<object>> request)
```
RaidMax: Per previous request, we should do the error handling in the RadarService not wrapped here.

RaidMax: Typically this is done by returning the result as a domain specific result with a success or failure indicator. Then the controller will handle converting the result to a status code rather than a blank try catch (exception handling should be done at the service layer)

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
```diff
@@ -0,0 +1,175 @@
...
+    public async Task<object> GetPlayersAsync(Server server, CancellationToken token)
+    {
+        var cache = _cache.GetOrAdd(server.ToString(), _ => new CacheEntry());
+        if (cache.PlayerExpiresAt > DateTime.UtcNow && cache.Players is not null)
+        {
+            return cache.Players;
+        }
+
+        var response = await server.ExecuteCommandAsync("listplayers", token);
+        cache.Players = SevenDaysToDiePlayerParser.Parse(string.Join(Environment.NewLine, response))
```
RaidMax: I'm a bit confused why we are executing commands against the server explicitly here. This becomes a separate process to the clients already tracked by IW4MAdmin via the regular status polling. A proper integration would be to reuse the client list already tracked by IW4MAdmin and cast them back to `SevenDaysToDiePlayer` so that we can access the additional properties

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
```diff
@@ -0,0 +1,175 @@
...
+
+public sealed class SevenDaysToDieRadarService(IHttpClientFactory httpClientFactory)
+{
+    private const int MaxTileBytes = 4 * 1024 * 1024;
+    private const int MaxSecretBytes = 4 * 1024;
+    private static readonly TimeSpan MapLifetime = TimeSpan.FromMinutes(1);
+    private static readonly TimeSpan PlayerLifetime = TimeSpan.FromSeconds(1);
+    private readonly ConcurrentDictionary<string, CacheEntry> _cache = new();
+
+    public async Task<object> GetMapAsync(Server server, CancellationToken token)
```
RaidMax: Returning Task<object> is bad practice. There should be a record or dto to encapsulate the fields.

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
```diff
@@ -0,0 +1,175 @@
...
+            saveName,
+            mapSize = new { x = worldSize, y = 255, z = worldSize },
+            tileSize = 128,
+            maxZoom = 4
+        };
+        cache.MapExpiresAt = DateTime.UtcNow + MapLifetime;
+        return cache.Map;
+    }
+
+    public async Task<object> GetPlayersAsync(Server server, CancellationToken token)
```
RaidMax: Returning Task\<object\> is bad practice. There should be a record or dto to encapsulate the fields.

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
```diff
@@ -0,0 +1,175 @@
...
+                throw new InvalidOperationException("ExternalWeb token configuration is incomplete or invalid");
+            }
+
+            var secretInfo = new FileInfo(configuration.TokenFile);
+            if (!secretInfo.Exists || secretInfo.Length is <= 0 or > MaxSecretBytes)
+            {
+                throw new InvalidOperationException("ExternalWeb token file is missing or invalid");
+            }
+
+            var secret = (await File.ReadAllTextAsync(configuration.TokenFile, token)).Trim();
```
RaidMax: Pretty certain this is going to be read every single time someone gets a tile. This should be cached and not loaded every time this method is called.

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
```diff
@@ -0,0 +1,175 @@
...
+
+    public async Task<object> GetMapAsync(Server server, CancellationToken token)
+    {
+        var cache = _cache.GetOrAdd(server.ToString(), _ => new CacheEntry());
+        if (cache.MapExpiresAt > DateTime.UtcNow && cache.Map is not null)
+        {
+            return cache.Map;
+        }
+
+        var world = await GetPreferenceAsync(server, "GameWorld", token) ?? "Unknown";
```
RaidMax: How often does this change? Do we need to get this _every_ single request?

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
```diff
@@ -0,0 +1,175 @@
...
+        using var response = await httpClientFactory.CreateClient("LiveRadar7DTD").SendAsync(request,
+            HttpCompletionOption.ResponseHeadersRead, token);
+        response.EnsureSuccessStatusCode();
+        if (!string.Equals(response.Content.Headers.ContentType?.MediaType, "image/png",
+                StringComparison.OrdinalIgnoreCase) || response.Content.Headers.ContentLength is > MaxTileBytes)
+        {
+            throw new InvalidDataException("External web endpoint did not return a valid PNG tile");
+        }
+
+        await using var input = await response.Content.ReadAsStreamAsync(token);
```
RaidMax: ASP.NET supports PipeWriter/Stream so that we can directly stream the response via the controller signature to the user instead of buffering the data. This increase efficiency and reduces memory footprint.

Plugins/LiveRadar/Services/SevenDaysToDieRadarService.cs
RaidMax: Alternatively it may make more sense to have this run on a scheduled timer instead of on demand so that the data is always fresh.

PR #413, opened 2026-09-07
cod4x IPV6 Support

xoxod33p:


Ayymoss:
Has this been checked? How does the database handle/store these IPv6 addresses? I don't recall if we can even support this without some more fundamental changes.

